PR #2, opened 2020-11-26
tests - openssl and python

briantist:
##### SUMMARY
Work on remaining CI issues

##### ADDITIONAL INFORMATION

Building on @dericcrago 's work from #1

PR #3, opened 2020-11-27
Try to make tests work.

felixfontein:


tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/main.yml
```diff
@@ -40,10 +40,16 @@
...
+    #
+    #- name: 'Install unzip'
+    #  package:
+    #    name: unzip
+    #  when: ansible_distribution != "MacOSX"  # unzip already installed
     - name: 'Install unzip'
-      package:
-        name: unzip
-      when: ansible_distribution != "MacOSX"  # unzip already installed
+      command: apt-get install unzip
```
briantist, line 52: I had the done the same (but in `runme.sh`) in #2 , but mostly as a workaround to get to the next error.

My main concern with this is that it forces the tests to only ever run on debian/ubuntu. I get that's all we're doing in CI now, not sure that will always be the case; and it kind of precludes anyone running the tests on their own outside of debian derivatives too, doesn't it?

Once the rest were working, I was thinking of going back and maybe adding a task conditional on debian/ubuntu that tries to implement some of the workarounds in this post: https://stackoverflow.com/a/53153996/3905079

tests/integration/targets/lookup_hashi_vault/playbooks/install_dependencies.yml
```diff
@@ -1,8 +1,8 @@
 - hosts: localhost
   tasks:
-    - name: Install openssl
-      import_role:
-        name: setup_openssl
+    - name: Install cryptography
+      pip:
+        name: cryptography
```
briantist, line 5: Was this actually needed? I expected to do it in #2 but before adding this in, I just took out all the openssl stuff and nothing seemed to break?

tests/integration/targets/lookup_hashi_vault/runme.sh
```diff
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 
+# The default container does not seem to have virtualenv installed
+${ANSIBLE_TEST_PYTHON_INTERPRETER} -m pip install virtualenv
+
 set -eux
 
-ansible-galaxy -vvv collection install community.crypto
+source virtualenv.sh
```
briantist, line 8: I did finally figure out that the issues were venv-based because of some combination of the ansible container and `ansible-test`'s use of virtual environments, but I've never heard of `virtualenv.sh`, is this documented anywhere at all??

This stuff was driving me crazy, thank you.

.github/workflows/ansible-test.yml
```diff
@@ -80,18 +80,26 @@ jobs:
...
         uses: actions/checkout@v2
         with:
           path: ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}
 
-      - name: Set up Python ${{ matrix.ansible }}
+      - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: ${{ matrix.python }}
+          python-version: 3.8
```
briantist, line 102: Should these be set back to the matrix values now?

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/main.yml
felixfontein: I think it will be a lot easier to fix properly once something like ansible/ansible#72430 is available. The `default` docker container is sufficiently different from the OS containers that could also be used. Once it is possible to distinguish them, one can use `command: apt-get install ...` for the default container and the `package` module in other cases.

@mattclay this is one case where knowing the target seems to be not enough, since there is no target. All tests are always run, the only difference is which values for `--docker` and `--python` are passed to `ansible-test`.

tests/integration/targets/lookup_hashi_vault/playbooks/install_dependencies.yml
felixfontein: Not really, because it's already installed in the `default` container. I wouldn't rely on that though, that's why I explicitly added it.

tests/integration/targets/lookup_hashi_vault/runme.sh
felixfontein: I have no idea idea where/whether it is documented. I mainly know that some `runme.sh` based tests use it (and need it), to make sure that the modules are executed with the correct Python version (which is not necessarily equal to the Python version with which ansible-playbook is run).

.github/workflows/ansible-test.yml
felixfontein: No. It should have never been the matrix value. (https://github.com/ansible-collections/collection_template/blob/main/.github/workflows/ansible-test.yml#L161)

briantist: Ohhh interesting, so how does the right python version get into place for `ansible-test` to use? Is it that the default container just has all of them?

briantist (approved):
thanks for this, lgtm

.github/workflows/ansible-test.yml
felixfontein: The default container has all of them. Or well, almost ;) Ansible 2.9's ansible-test only has Pythons up to Python 3.8, while 2.9 and devel also support Python 3.9. That's why there's the `exclude` entry in the matrix.

felixfontein:
CC @dericcrago @dmsimard could you review/merge please? :)

.github/workflows/ansible-test.yml
```diff
@@ -80,18 +80,26 @@ jobs:
...
         python:
+          - 2.7
+          - 3.5
           - 3.6
+          - 3.7
+          - 3.8
+          - 3.9
+        exclude:
+          - ansible: stable-2.9
+            python: 3.9
```
dmsimard, line 91: @felixfontein a comment to explain why 2.9 is excluded would be appreciated

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/main.yml
dmsimard: Echoing the concern about this failing on distros other than debian/ubuntu but we can live with this for now.

tests/integration/targets/lookup_hashi_vault/runme.sh
```diff
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 
+# The default container does not seem to have virtualenv installed
+${ANSIBLE_TEST_PYTHON_INTERPRETER} -m pip install virtualenv
```
dmsimard, line 4: Maybe add a link to https://github.com/ansible/ansible/issues/72738 in a comment ?

dmsimard (approved):

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/main.yml
mattclay: Why does the `package` module not work in this case?

briantist: @mattclay `package` tries to load the `apt` python module, which fails. It has a fallback for that, whereby it installs the required `python3-apt` system package then tries again. However what happens under the covers is that apt says "`python3-apt` is already installed", which it is. But if I understand it correctly, the reason it can't be loaded is that you need these apt bindings for every version of python, and the one installed is for the system version of python (which is 3.6), so in 3.6 runs it all works fine, but if the run uses any other version of python, it will fail, and there's no simple solution. It seems like what people usually do for this situation is symlink the installed library to the other python version folders. There are a number of such answers described in this SO question, here's an answer I thought explained it succinctly: https://stackoverflow.com/a/53153996/3905079

nitzmahone: This will obviously be a larger problem with the Python 3.8+ controller change that's slated for 2.11- that won't affect remotes, but modules local to the controller will inherit the controller Python for executing them, and run afoul of this issue. It's most noticeable on RHELish things that need selinux and/or yum/dnf Python libs, but there are other places (as you've found).

We've got several possible ways of handling this- the most likely candidate is "module respawn"- basically, for modules that know they need a weird library that's only available under a "system Python" like this, there'd be a standard module_util that allows the module to do whatever logic it wants to determine the right Python interpreter, then transparently respawn the module under that interpreter as a child process and pass the results through the original module process (which would just act as a shim to smuggle the results when running in respawn mode).

I had a working prototype of it a couple years ago (where it was trivial), but that was before AnsiballZ had switched to using runpy, which complicates a couple of the details of the respawn. It's a problem we have to solve for 2,11, anyway, though I doubt we'll be backporting it. Short-term, just manually setting the interpreter in a task var to the one that has the necessary libs for on-controller tests is probably the path of least resistance.

briantist: thanks @nitzmahone , I think I'll take a stab at doing that (in a new PR): putting the `package` task back but passing in a desired "system" python to use on that task.

briantist: For anyone who wants to follows along: https://github.com/ansible-collections/community.hashi_vault/pull/6

The above change seemed to work fine for Python 3, in 2.7 the tests all run but fail at the end complaining that the tests modified the environment. Will look more into that.

PR #11, opened 2020-12-02
Release 0.1.0

briantist:
##### SUMMARY
Release 0.1.0

##### ISSUE TYPE

Release

### DIFF
--- a/changelogs/changelog.yaml
+++ b/changelogs/changelog.yaml
@@ -3,7 +3,7 @@ releases:
   0.1.0:
     changes:
       release_summary: 'Our first release matches the ``hashi_vault`` lookup functionality
-        provided by ``community.general`` ver ``1.3.0``.
+        provided by ``community.general`` version ``1.3.0``.
 
 '
     fragments:
--- a/changelogs/changelog.yaml
+++ b/changelogs/changelog.yaml
@@ -1,1 +1,11 @@
-releases: {}
+ancestor: null
+releases:
+  0.1.0:
+    changes:
+      release_summary: 'Our first release matches the ``hashi_vault`` lookup functionality
+        provided by ``community.general`` ver ``1.3.0``.
+
+'
+    fragments:
+    - 0.1.0.yml
+    release_date: '2020-12-02'
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -7,13 +7,15 @@ readme: README.md
 authors:
   - Jonathan Davila (!UNKNOWN) <jdavila(at)ansible.com>
   - Brian Scholer (@briantist)
-description: null
+description: Plugins related to HashiCorp Vault
 license_file: LICENSE
 tags:
   # tags so people can search for collections https://galaxy.ansible.com/search
   # tags are all lower-case, no spaces, no dashes.
-  - example1
-  - example2
+  - vault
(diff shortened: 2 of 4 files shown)


changelogs/changelog.yaml
```diff
@@ -1 +1,11 @@
-releases: {}
+ancestor: null
+releases:
+  0.1.0:
+    changes:
+      release_summary: 'Our first release matches the ``hashi_vault`` lookup functionality
+        provided by ``community.general`` ver ``1.3.0``.
+
+'
```
dmsimard, line 8: @felixfontein do you know where that odd extra whitespace comes from ? I've noticed that in other relnotes as well

dmsimard (approved):
LGTM, note that we must merge the zuul PRs before actually tagging the release

changelogs/changelog.yaml
felixfontein: That's a YAML representation of https://github.com/ansible-collections/community.hashi_vault/pull/11/commits/1f53ba788f9ecbdc9799568133538d1a4455eb00#diff-564fa06adc205ddea56f8b8154bdb228ee47a7904bb469f56ec47f17a43f83a7

changelogs/changelog.yaml
```diff
@@ -1 +1,11 @@
-releases: {}
+ancestor: null
+releases:
+  0.1.0:
+    changes:
+      release_summary: 'Our first release matches the ``hashi_vault`` lookup functionality
+        provided by ``community.general`` ver ``1.3.0``.
```
felixfontein, line 6: I think `version` should be written out.

galaxy.yml
```diff
@@ -7,13 +7,15 @@ readme: README.md
 authors:
   - Jonathan Davila (!UNKNOWN) <jdavila(at)ansible.com>
```
felixfontein, line 8:
```suggestion
  - Jonathan Davila
```
That specific format is needed for plugins/modules, but not here :)

changelogs/changelog.yaml
briantist: er, I do too, I must've gotten distracted in the middle of writing the word 🤦‍♂️

changelogs/changelog.yaml
dmsimard: Ah so extra whitespace because of the pipe ``|`` ? Would ``>-`` work better for example ?

changelogs/changelog.yaml
felixfontein: Hehe, that does happen :)

changelogs/changelog.yaml
felixfontein: It depends on whether you plan to have multiple lines or even paragraphs. If you use `>` instead of `|`, it will be really hard to create new paragraphs since you need two newlines in the parsed string, and a double newline in YAML only gives you one newline in the parsed string. Probably `|-` would be better; I've adjusted https://github.com/ansible/community/wiki/ReleasingCollections accordingly.

briantist: Should I adjust or remove the extra generated whitespace or just leave it?

felixfontein (approved):

dericcrago (approved):

Issue #10, opened 2020-12-01
hashi_vault - environment variable guideline proposal


briantist:
##### SUMMARY
The `hashi_vault` lookup takes several parameters via environment variables, and recently, some via INI (`ansible.cfg`) as well.

Most of the original vars were intended to re-use [vars that are commonly used in the vault CLI.](https://www.vaultproject.io/docs/commands#environment-variables), like `VAULT_ADDR` and `VAULT_TOKEN`.

Along the way some newer ones have been added that aren't related to the vault CLI but are strictly for the plugin, but also adopted the `VAULT_` prefix, and some others used `ANSIBLE_VAULT_` as their prefix.

An additional consideration is that in Ansible, environment variables have higher precedence (are preferred) over INI values. We can think of this in terms of specificity; the INI value is defined "ansible-wide" (for your specific implementation), while the env var is defined _per invocation_ so it should be considered more specific, **especially given that the env vars defined for plugin parameters are specific to that Ansible plugin** (usually denoted with the `ANSIBLE_` prefix) and so have no use outside of that context.

The precedence issue becomes less clear when we take values that have meaning outside of ansible, like the `VAULT_` vars used in the vault CLI.

I believe we should consider then that an env var with meaning outside of Ansible has a _lower precedence_ (is _less preferred_) than an INI value defined in `ansible.cfg`, as in the context of Ansible, that is a more specific value.

##### PROPOSAL

This proposal aims to provide guidelines on how we name environment variables, and how we handle their precedence.

###### Ansible-only vars
This applies to a parameter to the plugin that will not inherit any existing vault CLI env var.

1. **The variable should use the prefix `ANSIBLE_HASHI_VAULT_`**. This name is on the long side, however it keeps to the pattern of starting with `ANSIBLE_`. Also considered (and even used before this proposal) was `ANSIBLE_VAULT_` however this may cause confusion with [Ansible Vault](https://docs.ansible.com/ansible/latest/user_guide/vault.html) (unrelated to this plugin).
1. This variable should follow existing Ansible precedence rules, and be specified in the `env:` key in the plugin's argspec, so that the standard plugin loader can take care of it.

Even if there is no specific existing `VAULT_` env var, we don't know whether one will be introduced in the future, and so don't want to use these names.

###### Variables used outside of Ansible
[...]

PR #21, opened 2020-12-09
hashi_vault - internal changes for env var processing and unwanted token loading fix

### LINKED ISSUE #13: hashi_vault - token used for auth may inadvertently be inferred
##### SUMMARY
See: https://github.com/ansible-collections/community.general/pull/1213#pullrequestreview-530759779

What we found during the test run in the linked PR is that HVAC was using the root token of our Vault dev server used in CI. This masked an issue with the JWT auth method in HVAC not behaving like other auth methods ( https://github.com/hvac/hvac/issues/644 ) because we ended up using the root token instead.

The only test that caught this was the one that expected an access denied error with the given auth, because of course the root token had permission. If that test had not been there, it would have gone unnoticed and JWT auth would have been completely broken, with passing tests.

In https://github.com/ansible-collections/community.general/pull/23 this bug was fixed for the general case of a failed auth method using an otherwise implied token, but it seems the HVAC library itself is doing some inference that we haven't accounted for. This may be specific to Vault being in dev mode.

---

Proposed investigation steps:
* Investigate the HVAC codebase to determine how it's looking for implicit tokens
* Look at what conditions this happens under
* Plan further resolution

Possible remediation options:
* replicate all implicit token loading as first-class options, the way we have for `VAULT_TOKEN` and the token file created by `vault login`
* perform an immediate `.logout()` after we first create the HVAC client, ensuring we don't have anything implicit that we haven't loaded via the plugin
* come up with a better test for this scenario, hopefully one that isn't specific to any one auth method, to quickly catch regressions

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

hashi_vault.py

briantist:
##### SUMMARY
Most of this PR sets the stage for addressing the issues related to #10 and tracked in https://github.com/ansible-collections/community.hashi_vault/projects/1 .

Most of those will be addressable now by adding new entries to the `LOW_PRECEDENCE_ENV_VAR_OPTIONS` var if they need such a thing (the rest are just argspec `env:` replacements).

The change here sets that up, and moves that functionality for `token_path`'s `$HOME` env var into that (as it's the only option already using such a thing). As such it doesn't change the functionality or user interface, just the implementation.

This PR also fixes #13 by forcing the hvac client to logout when not doing token auth. 
This also doesn't change any current functionality, it just helps prevent auth methods from inadvertently not assigning their token to the client (previously that could have been masked by an accidentally inferred token, which is a problem in regular environments as well as in our CI).

Couple of minor comment/doc updates thrown in too.

No changelog since there are no observable changes for users.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

hashi_vault.py

### DIFF
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -241,6 +241,8 @@ EXAMPLES = """
   ansible.builtin.debug:
     msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/hello:value', auth_method='aws_iam_login', role_id='myroleid', profile=my_boto_profile) }}"
 
+# JWT auth
+
 - name: Authenticate with a JWT
   ansible.builtin.debug:
     msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/hola:val', auth_method='jwt', role_id='myroleid', jwt='myjwt', url='https://vault:8200') }}"
@@ -283,6 +285,20 @@ try:
     HAS_BOTO3 = True
 except ImportError:
     HAS_BOTO3 = False
+
+# see https://github.com/ansible-collections/community.hashi_vault/issues/10
+#
+# Options which seek to use environment vars that are not Ansible-specific
+# should load those as values of last resort, so that INI values can override them.
+# For default processing, list such options and vars here.
+# Alternatively, process them in another appropriate place like an auth method's
+# validate_ method.
+#
+# key = option_name
+# value = list of env vars (in order of those checked first; process stops when value is found)
+LOW_PRECEDENCE_ENV_VAR_OPTIONS = {
+    'token_path': ['HOME'],
+}
 
 
 class HashiVault:
@@ -393,7 +409,7 @@ class HashiVault:
     #    already have everything they need. This is also the place to check for deprecated auth methods as hvac
     #    continues to move backends into the auth_methods class.
     #
-    # 3. Update the avail_auth_methods list in the LookupModules auth_methods() method (for now this is static).
+    # 3. Update the avail_auth_methods list in the LookupModule's auth_methods() method (for now this is static).
     #
     def auth_token(self):
         if not self.client.is_authenticated():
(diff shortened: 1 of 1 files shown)


plugins/lookup/hashi_vault.py
```diff
@@ -318,6 +334,10 @@ def __init__(self, **kwargs):
             client_args['token'] = self.options.get('token')
 
         self.client = hvac.Client(**client_args)
+        # logout to prevent accidental use of inferred tokens
+        # https://github.com/ansible-collections/community.hashi_vault/issues/13
+        if 'token' not in client_args:
+            self.client.logout()
```
felixfontein, line 340: I think this requires I changelog fragment.

felixfontein (commented):
Looks good! One of the changes probably needs a changelog fragment, since it could affect users.

plugins/lookup/hashi_vault.py
briantist: @felixfontein I don't think it does but certainly willing to hear you out! 

This doesn't change current behavior, because every auth method (except token), sets the token in the client after this point (all of them do it implicitly except JWT, for that one we do it explicitly in this plugin; that's an HVAC inconsistency).

For token auth, we set that before creating the client, so we don't call logout.

Calling `.logout()` here is to ensure that the token is clear before we continue to the auth methods, because when a token is not supplied during client creation, it tries to find one (in env or in token file). In all normal cases that doesn't matter: the auth  method will overwrite it.

So what this really does is prevent a developer from accidentally not ensuring a token is set in a new auth method (or when modifying an existing one). Before this change, what would happen is that the client still had its inferred token, so it would use it in further operations, which would give a confusing result. 

That's especially true in our CI where a root token with full privileges is available. This was surfaced in the PR to add JWT auth, when all the tests were passing except the one that tried to access a secret it was not supposed to have access to (expecting access denied), and inexplicably it was able to access the secret.

It was seemingly so weird we were considering whether it was a bug in vault and whether to remove the test (which would have been a disaster because in truth the auth method was never actually setting the client with the token it received from doing JWT auth).

--- 

So in short, all existing auth methods would never trigger this, this change is to prevent future implementations from passing a bunch of tests by accident because of re-used tokens.

felixfontein (approved):

PR #25, opened 2020-12-12
hashi_vault - parameter env var changes for token and namespace

### LINKED ISSUE #14: hashi_vault - [env] Update VAULT_NAMESPACE to match guidelines & add INI
##### SUMMARY
See #10 

`VAULT_NAMESPACE` is an environment variable used by the Vault CLI. I'm proposing the following change:

* Add `ANSIBLE_HASHI_VAULT_NAMESPACE` to `env:` in spec
* Add INI value as well
* Process `VAULT_NAMESPACE` as a fallback to populate the option when it hasn't been set via any other means.

ℹ This change is not breaking because lowering the precedence of the existing env var doesn't raise any existing value's precedence (there's no other env var or INI value currently).

##### ISSUE TYPE
- Feature Request

##### COMPONENT NAME

hashi_vault.py

### LINKED ISSUE #16: hashi_vault - [env] Update VAULT_TOKEN handling
##### SUMMARY
See #10 

`VAULT_TOKEN` is a Vault CLI env var.

The proposal then is to add an Ansible-specific env var and lower the precedence of the existing one:
* Add `ANSIBLE_HASHI_VAULT_TOKEN` in `env:` in spec
* Process `VAULT_TOKEN` as a fallback if the option is not set directly or via spec-defined env
* 🎗this parameter already has another fallback action, which is to try to infer its value via the `token_file` and `token_path` parameters. This will remain, and it will still have lower precedence than the `VAULT_TOKEN` env var. This preserves the existing relative relationship.

ℹThis is not a breaking change
* existing ways of specifying the value keep their relative precedence, and the new env var didn't exist previously

##### ISSUE TYPE
- Feature Idea

briantist:
##### SUMMARY
Resolves #14 
- Add `ANSIBLE_HASHI_VAULT_NAMESPACE` env var to argspec
- Move `VAULT_NAMESPACE` env var to `LOW_PRECEDENCE_ENV_VAR_OPTIONS`
- Add new INI option for `namespace`

Resolves #16 
- Add `ANSIBLE_HASHI_VAULT_TOKEN` env var in argspec
- Move `VAULT_TOKEN` env var to `LOW_PRECEDENCE_ENV_VAR_OPTIONS`

These changes are non-breaking (see issues for detailed explanation).

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

hashi_vault.py

##### ADDITIONAL INFORMATION
Overall project: https://github.com/ansible-collections/community.hashi_vault/projects/1
See also: #10

### DIFF
--- a/changelogs/fragments/25-non-breaking-env-parameter-changes.yml
+++ b/changelogs/fragments/25-non-breaking-env-parameter-changes.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - hashi_vault - ``namespace`` parameter can be specified in INI or via env vars ``ANSIBLE_HASHI_VAULT_NAMESPACE`` (new) and ``VAULT_NAMESPACE`` (lower preference)  (https://github.com/ansible-collections/community.hashi_vault/issues/14).
+  - hashi_vault - ``token`` parameter can now be specified via ``ANSIBLE_HASHI_VAULT_TOKEN`` as well as via ``VAULT_TOKEN`` (the latter with lower preference).
--- a/changelogs/fragments/25-non-breaking-env-parameter-changes.yml
+++ b/changelogs/fragments/25-non-breaking-env-parameter-changes.yml
@@ -1,4 +1,4 @@
 ---
 minor_changes:
   - hashi_vault - ``namespace`` parameter can be specified in INI or via env vars ``ANSIBLE_HASHI_VAULT_NAMESPACE`` (new) and ``VAULT_NAMESPACE`` (lower preference)  (https://github.com/ansible-collections/community.hashi_vault/issues/14).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/25-non-breaking-env-parameter-changes.yml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
+  - hashi_vault - ``namespace`` parameter can be specified in INI or via env vars ``ANSIBLE_HASHI_VAULT_NAMESPACE`` (new) and ``VAULT_NAMESPACE`` (lower preference)  (https://github.com/ansible-collections/community.hashi_vault/issues/14).
+  - hashi_vault - ``token`` parameter can now be specified via ``ANSIBLE_HASHI_VAULT_TOKEN`` as well as via ``VAULT_TOKEN`` (the latter with lower preference).
```
felixfontein, line 4: There's no URL for this changelog entry.

felixfontein (approved):

PR #26, opened 2020-12-13
CI/Tests - Caching, multiple vault versions, etc.

briantist:
##### SUMMARY

**Tests**
- Reorganized the parts of tests dedicated to downloading Vault and starting a Vault server to be more discrete
- Vault download / extraction happens in the role's `files/bin` directory now (`.gitignored`), more on that later
- Download is skipped if the requested version of the binary already exists, saving the download and extraction steps
- Playbook added that calls just the Vault download portion of the role
- Vars related to vault version and location are moved to `defaults/` to better control overriding them and so that they are available to all role entry points
- Download Vault binaries directly from hashiCorp's releases server
- "Default" version of Vault to be tested against is now 1.6.0, the current latest release.

**CI**
- Some deduping of vars and paths by setting them in env vars in a job step (integration only)
- Add Vault version to Matrix
  - With vars changes above, we can write a `vars/main` that overrides the Vault version set in `defaults/main`, controlling the version for everything invoked in CI
  - With download changes above, we can call the vault download playbook in a separate step to download the vault binary before starting the integration test (this will also implicitly use the vault version we wrote)
- Using the `cache` action, we can tell GitHub to cache the contents of the `files/bin` directory, based on a key we build from the runner OS and the Vault version. if the key is found, the cached contents are loaded into the directory.
  - With a cache hit, we skip the pre-download job altogether (if we run it anyway, it still won't re-download or extract anything because the role is written that way)
- Because `.gitignore` files were deleted, `ansible-test` will send the downloaded/cache-restored contents
- The `ansible-test` run is otherwise oblivious to the caching, but the file will exist, and so it will not re-download or extract anything

---

The caching isn't going to save us _that_ much time overall, although I will say the longest part of the process seemed to be installing `unzip` into Ubuntu 😕and that's now skipped generally. However the caching should help smooth over any issues with hosted binaries, as in the vast majority of runs we will be pulling from cache.

Cache entries are evicted after not being accessed for 7 days. If things slow enough on the repo that  no tests get run in that long, we can either decide that's fine, or set a cron trigger in CI to every 6 days or so to keep the cache primed.

---

**Running Tests Locally**
Using `ansible-test` locally still works like before; devs don't have to do anything special. But there are some caveats:
- If using `--docker` (as I do, and generally recommend everyone do), you won't be able to pre-download the binary or override the version with the `vars/` folder, because of the contents being `.gitignored`.
- You can delete the ignore files locally, but then you'll have unwanted git changes to deal with, which is also not ideal, but it is doable.

In any case, local `ansible-test` is fully functional, just not quite as optimized unless you're ok with being careful about your git changes

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -23,10 +23,6 @@ jobs:
 # https://docs.ansible.com/ansible/latest/dev_guide/testing_sanity.html
 
   sanity:
-    env:
-      COLLECTION_PATH: ${{ needs.vars.outputs.collection_path }}
-      LOOKUP_HASHI_VAULT_PATH: ${{ needs.vars.outputs.lookup_hashi_vault_path }}
-      LOOKUP_HASHI_VAULT_BIN: ${{ needs.vars.outputs.lookup_hashi_vault_bin }}
     name: Sanity (Ⓐ${{ matrix.ansible }})
     strategy:
       matrix:
@@ -61,7 +57,7 @@ jobs:
       # and all python versions ansible supports.
       - name: Run sanity tests
         run: ansible-test sanity --docker -v --color
-        working-directory: ./ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}
+        working-directory: ./ansible_collections/${{ env.NAMESPACE }}/${{ env.COLLECTION_NAME }}
 
 ###
 # Integration tests (RECOMMENDED)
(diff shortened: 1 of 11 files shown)


.github/workflows/ansible-test.yml
```diff
@@ -86,36 +87,76 @@ jobs:
...
 
     steps:
+      - name: Initialize env vars
+        run: |
+          COLLECTION_PATH=ansible_collections/${NAMESPACE}/${COLLECTION_NAME}
+          LOOKUP_HASHI_VAULT_PATH=${COLLECTION_PATH}/tests/integration/targets/lookup_hashi_vault
+          LOOKUP_HASHI_VAULT_BIN=${LOOKUP_HASHI_VAULT_PATH}/lookup_hashi_vault/files/bin
+          LOOKUP_HASHI_VAULT_VARS=${LOOKUP_HASHI_VAULT_PATH}/lookup_hashi_vault/vars
+
+          echo "COLLECTION_PATH=${COLLECTION_PATH}" >> $GITHUB_ENV
```
felixfontein, line 106:
```suggestion
          echo "COLLECTION_PATH=${COLLECTION_PATH}" >> ${GITHUB_ENV}
```
(same for the next ones?)

.github/workflows/ansible-test.yml
```diff
@@ -86,36 +87,76 @@ jobs:
...
+          rm -f "${LOOKUP_HASHI_VAULT_VARS}/.gitignore"
+          echo '{ "vault_version": "${{ matrix.vault }}" }' > "${LOOKUP_HASHI_VAULT_VARS}/main.json"
+
       - name: Install ansible-base (${{ matrix.ansible }})
         run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check
 
       - name: Install collection dependencies
         run: ansible-galaxy collection install community.crypto -p .
 
+      # this wil populate files/bin with the selected vault version binary
```
felixfontein, line 143:
```suggestion
      # this will populate files/bin with the selected vault version binary
```

.github/workflows/ansible-test.yml
```diff
@@ -86,36 +87,76 @@ jobs:
...
+          - 1.4.7
         exclude:
           - ansible: stable-2.9
             python: 3.9
 
     steps:
+      - name: Initialize env vars
+        run: |
+          COLLECTION_PATH=ansible_collections/${NAMESPACE}/${COLLECTION_NAME}
+          LOOKUP_HASHI_VAULT_PATH=${COLLECTION_PATH}/tests/integration/targets/lookup_hashi_vault
```
felixfontein, line 102: It would make sense to move the setup part to a different target `setup_hashi_vault`, so that once you add more modules / plugins you can simply depend on `setup_hashi_vault` as well. (But you can also do that split-up in a later PR.)

briantist: I totally agree, but it will later/in a different PR. There's more I'd like to do as well, like not just being able to set a vault version, but also supply the address of a running instance, in which case download/extract/set policies and secrets, etc. can all be skipped.

But yeah separating it out later will let the whole thing get leaner. I just didn't want to add even more to what are already a lot of changes.

felixfontein (approved):

Issue #7, opened 2020-10-30
lookup/hashi_vault.py: IAM auth does not support the `mount_point` parameter


lamawithonel:
##### SUMMARY
AWS IAM authentication does not take the `mount_point` parameter, but other auth methods do.

##### ISSUE TYPE
- Bug Report

A previous report of a similar issue (#767) was classified as a Bug.

##### COMPONENT NAME
`lookup/hashi_vault.py`

##### ANSIBLE VERSION
```
ansible 2.10.2
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.2 (default, Jul 18 2020, 19:35:03) [GCC 9.2.0]
```

##### CONFIGURATION
No changes

##### OS / ENVIRONMENT
```sh
cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.11.6
PRETTY_NAME="Alpine Linux v3.11"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"
```

##### STEPS TO REPRODUCE

1. Enable the AWS auth method in HashiCorp Vault with a custom mountpoint, e.g., `obfuscated-iam`.
2. Use the `hash_vault` lookup plugin with the options `auth_method=aws_iam_login mount_point=obfuscated-iam`

##### EXPECTED RESULTS

`hashi_vault` should authenticate successfully to Vault using the custom mountpoint.

The issue resides, or at least is fixable, in the `validate_auth_aws_iam_login` function, [here](https://github.com/ansible-collections/community.general/blob/3af4be34b230af74d47f9c9411d037da08357f4a/plugins/lookup/hashi_vault.py#L585-L618).  A naive fix is to add `mount_point` to the `params` list as below:

```python
    def validate_auth_aws_iam_login(self, auth_method):
        params = {
            'access_key': self.get_option('aws_access_key'),
            'secret_key': self.get_option('aws_secret_key'),
            'mount_point': self.get_option('mount_point')   # <--------- a naive fix
        }
        ...
```

##### ACTUAL RESULTS
[...]

PR #31, opened 2020-12-19
hashi_vault - refactor tests with mount_point capabilities. fix jwt mount_point options

### LINKED ISSUE #29: Parameter 'mount_point' does not work with JWT auth
##### SUMMARY
The parameter `mount_point` does not work with JWT auth. It seems like the parameter must be renamed/aliased to `path` before invoking `hvac.jwt_login()`. Ref. [hvac.jwt_login() documentation](https://hvac.readthedocs.io/en/stable/source/hvac_api_auth_methods.html#hvac.api.auth_methods.JWT.jwt_login). I think this can be fixed quite easily with the same approach as used for aliasing the Vault role parameter: https://github.com/ansible-collections/community.hashi_vault/blob/a4374fd904f1b61b3eb7f108cfd89a96b5c76054/plugins/lookup/hashi_vault.py#L472

But adding tests for this seems like a considerable job....

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
`lookup/hashi_vault.py`

##### ANSIBLE VERSION

```paste below
ansible 2.10.4
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.9.1 (default, Dec 12 2020, 13:15:12) [GCC 8.3.0]
```

##### CONFIGURATION

```paste below
ANSIBLE_FORCE_COLOR(env: ANSIBLE_FORCE_COLOR) = True
DEFAULT_LOAD_CALLBACK_PLUGINS(env: ANSIBLE_LOAD_CALLBACK_PLUGINS) = True
DEFAULT_STDOUT_CALLBACK(env: ANSIBLE_STDOUT_CALLBACK) = yaml
INTERPRETER_PYTHON(env: ANSIBLE_PYTHON_INTERPRETER) = /usr/local/bin/python
```

##### OS / ENVIRONMENT

```
cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

##### STEPS TO REPRODUCE

```yaml
- name: Authenticate with a JWT using custom mountpoint/path
  ansible.builtin.debug:
      msg: "{{ lookup('community.general.hashi_vault', 'secret/hello:value', auth_method='jwt', role_id='myroleid', jwt='myjwt', url='https://myvault:8200', mount_point='mypath')}}"
```

##### EXPECTED RESULTS

`hashi_vault` should authenticate successfully to Vault using the custom mountpoint/path and debug the secret.

##### ACTUAL RESULTS

```paste below
Error was a <class ''TypeError''>, original message: jwt_login() got an unexpected keyword argument ''mount_point'''
```

briantist:
##### SUMMARY

Fixes #29 

- Refactors tests to include testing for `mount_point` in a few ways:
  - First, each auth method is added to the test vault server twice, once with the default mount path, and once with an alternate path
  - The tests are set up now to be run through twice. once with each path
  - When testing the default path, the `mount_point` parameter is specifically left out of the calls to the lookup; this will catch a situation where the default path that's set via the `vault` CLI (when we set up the method) and the default path used by `hvac` (when we use it in the plugin), differ. It's an unlikely event, and one that isn't quite our place to fix but something we should be aware of.
  - More importantly, in the alt mount cycle, we do use the `mount_point` parameter. This ensures that the parameter is being sent into `hvac` when supplied. This catches the situation in #29 where we thought we were passing the parameter in but had the wrong name (in this case due to JWT doing something different https://github.com/hvac/hvac/issues/655)
  - This leaves one area that wouldn't be caught by the above two, which is the one described in #7 : in that case, the `mount_point` parameter in the plugin was completely ignored. This results in the request being sent to vault with no `mount_point`, so using the default path. Since we just duplicate the policy in our test environment though, that would result in "alternate" mount point tests all being successful in this scenario, because under the hood they are using the default auth method mount which has all the permissions needed.
  - To catch that situation, I'm now creating two policies, with a new secret, `secret5`. The main test policy is denied access to that secret while the alternate policy has all the same access as the main one except that it can also read `secret5`. A new test was added to each auth method trying to access `secret5` and expecting failure on the default mount while expecting success otherwise.
  - #7 is a little bit special since we don't test the AWS auth method at all, but I'll introduce a new PR after this one that attempts to do some basic tests on the auth method (without actually interacting with AWS) to see what we can put in place, including trying to catch something like that.
- As for the actual fix for the JWT method, that was of course the easiest part. Although I've run through the tests locally while developing this, I'm putting up the PR first without the JWT fix commit, expecting the new tests to catch the JWT error. I'll push that commit afterwards which should cause it to pass.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

hashi_vault.py


briantist:
- New test suite catching #29:  https://github.com/ansible-collections/community.hashi_vault/pull/31/checks?check_run_id=1583019989#step:10:10910
-

briantist:
@erikgb might you have a chance to review and/or test this?

felixfontein (approved):

PR #34, opened 2020-12-21
hashi_vault - update env vars to align with standardization project

### LINKED ISSUE #8: hashi_vault - [env] INI value for vault URL should take precedence over VAULT_ADDR env var
##### SUMMARY

This is similar to https://github.com/ansible-collections/community.general/issues/373

We take the value of `VAULT_ADDR` in the plugin to emulate the vault CLI and use the env var, which is commonly set. In ansible-collections/community.general#23 an INI entry was added, but if `VAULT_ADDR` is set the INI value won't be used.

Since `VAULT_ADDR` is a more general value, in the context of Ansible, we should prefer the INI value if both are set.

The fix would be similar to that in ansible-collections/community.general#902 , manually loading the value. Also adding an additional (spec defined) env var, `ANSIBLE_HASHI_VAULT_ADDR` which could override the INI.

General env var guidance for this plugin: #10

**Plan:**
- `ANSIBLE_HASHI_VAULT_ADDR` was introduced in [v0.2.0](https://github.com/ansible-collections/community.hashi_vault/milestone/1)
- `VAULT_ADDR` will have is precedence lowered in [v1.0.0](https://github.com/ansible-collections/community.hashi_vault/milestone/2)

⚠This is _technically_ a **breaking** change
📝Impact Notes:
* It is unlikely to break most existing configurations, but it is possible. **All** of the following conditions must be met for a configuration to break:
  * The `url` parameter is **not** being passed explicitly to the plugin in the `lookup()` call
  * `url` is set in **both** the INI file **and** via the `VAULT_ADDR` env var
  * the values of the url in the above places are *different*
* With those conditions, the value of `VAULT_ADDR` is used before this change, and the other different value set in INI is used after this change.
* For users who are aware of this change before they consume it, it can be worked around in advance via the following remediations:
  * If the mismatched env and INI values are unintentional, setting them to be the same (or choosing one place to set them) can be done with no ill effects before and after the change.
  * If the mismatch is intentional (in one environment the INI value is used, while in another invocation the env var is being used to intentionally override INI), then this can be worked around: 
    * In `community.hashi_vault` `v0.1.0`: ideally upgrade to `v0.2.0` then follow the next step, otherwise: set both `VAULT_ADDR` and the not-yet-used `ANSIBLE_HASHI_VAULT_ADDR` to the same value in those invocations. After updating to a post-change collection (either `v0.2.0` or `v1.0.0`), remove the `VAULT_ADDR` setting.
    * In `community.hashi_vault` `v0.2.0`: set `ANSIBLE_HASHI_VAULT_ADDR` _instead of_ `VAULT_ADDR`. No other changes are needed for `v1.0.0`.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

hashi_vault.py

##### EXPECTED RESULTS

INI value is specific to ansible, and should be used over the more system wide `VAULT_ADDR`.

##### ACTUAL RESULTS

INI value will never be used if `VAULT_ADDR` is set.

### LINKED ISSUE #15: hashi_vault - [env] Fix token_file and token_path env vars
##### SUMMARY
See #10 

In https://github.com/ansible-collections/community.general/pull/902 I replaced/added env vars prefixed by `VAULT_` to solve the underlying issue, however thinking about that after the fact has led to #10 .

I'm proposing the following:
* The `VAULT_TOKEN_FILE` and `VAULT_TOKEN_PATH` env vars will be deprecated in [v0.2.0](https://github.com/ansible-collections/community.hashi_vault/milestone/1) and removed in [v2.0.0](https://github.com/ansible-collections/community.hashi_vault/milestone/3).
* The `ANSIBLE_HASHI_VAULT_TOKEN_FILE` and `ANSIBLE_HASHI_VAULT_TOKEN_PATH` vars will be added in [v0.2.0](https://github.com/ansible-collections/community.hashi_vault/milestone/1)
* The `VAULT_` variants will not be preserved, as they are not currently in use by the Vault CLI for this (or any) purpose; future use would conflict and this should be avoided.

⚠This is a **breaking** change
📝Impact Notes:
* Affects users who use `VAULT_` prefixed environment variables which are available in the following collections and versions:
  * `community.general` version `0.2.0` and above
  * `community.hashi_vault` version `0.1.0`
* Does not affect anyone using INI or direct parameter setting

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

hashi_vault.py

### LINKED ISSUE #17: hashi_vault - [env] update VAULT_AUTH_METHOD to match guidelines
##### SUMMARY
See #10 

`VAULT_AUTH_METHOD` uses the `VAULT_` prefix.

Proposed change:
* `ANSIBLE_HASHI_VAULT_AUTH_METHOD` will be added in [v0.2.0](https://github.com/ansible-collections/community.hashi_vault/milestone/1)
* `VAULT_AUTH_METHOD` will be deprecated in [v0.2.0](https://github.com/ansible-collections/community.hashi_vault/milestone/1) and removed in [v2.0.0](https://github.com/ansible-collections/community.hashi_vault/milestone/3)
* `VAULT_AUTH_METHOD` will not be preserved, as it is not currently in use by the Vault CLI for this (or any) purpose; future use would conflict and this should be avoided.

⚠This is a **breaking** change
📝Impact Notes:
* Affects users who use `VAULT_AUTH_METHOD` which is available in the following collections and versions:
  * `community.general` version `0.2.0` and above
  * `community.hashi_vault` version `0.1.0`

Does not affect anyone using INI or direct parameter setting.

##### ISSUE TYPE
- Feature Idea

### LINKED ISSUE #20: hashi_vault - [env] update role_id and secret_id env vars to match guidelines
##### SUMMARY
See #10

`VAULT_ROLE_ID` and `VAULT_SECRET_ID` use the `VAULT_` prefix.

Proposed change:
* `ANSIBLE_HASHI_VAULT_ROLE_ID` and `ANSIBLE_HASHI_VAULT_SECRET_ID` will be added in [v0.2.0](https://github.com/ansible-collections/community.hashi_vault/milestone/1)
* `VAULT_ROLE_ID` and `VAULT_SECRET_ID` will be deprecated in [v0.2.0](https://github.com/ansible-collections/community.hashi_vault/milestone/1) and removed in [v2.0.0](https://github.com/ansible-collections/community.hashi_vault/milestone/3)
* The `VAULT_` prefixed vars will not be preserved, as they are not currently in use by the Vault CLI for this (or any) purpose; future use would conflict and this should be avoided.

⚠This is a **breaking** change
📝Impact Notes:
* Affects users who use these vars (for the `approle`, `aws_iam_login`, and `jwt` auth methods) which are available in the following collections and versions:
  * `community.general` version `0.2.0` and above
  * `community.hashi_vault` version `0.1.0`

Does not affect anyone using INI or direct parameter setting.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
hashi_vault.py

briantist:
##### SUMMARY
This PR includes breaking env changes described in #10 and in the [Environment Variable Standardization](https://github.com/ansible-collections/community.hashi_vault/projects/1) project.

These changes are slated for [v1.0.0](https://github.com/ansible-collections/community.hashi_vault/milestone/2).

- Resolves #8 
- Resolves #17
- Resolves #15 
- Resolves #20 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

hashi_vault.py

### DIFF
--- a/changelogs/fragments/34-env-var-changes.yml
+++ b/changelogs/fragments/34-env-var-changes.yml
@@ -1,0 +1,8 @@
+---
+breaking_changes:
+  - hashi_vault - ``url`` parameter's env var ``VAULT_ADDR`` now has lower precedence than INI value (https://github.com/ansible-collections/community.hashi_vault/issues/8).
+  - hashi_vault - ``auth_method`` parameter's env var ``VAULT_AUTH_METHOD`` has been replaced by ``ANSIBLE_HASHI_VAULT_AUTH_METHOD`` (https://github.com/ansible-collections/community.hashi_vault/issues/17).
+  - hashi_vault - ``token_file`` and ``token_path`` parameters' env vars ``VAULT_TOKEN_FILE`` and ``VAULT_TOKEN_PATH`` have been replaced by ``ANSIBLE_HASHI_VAULT_TOKEN_FILE`` and ``ANSIBLE_HASHI_VAULT_TOKEN_PATH`` (https://github.com/ansible-collections/community.hashi_vault/issues/15).
+  - hashi_vault - ``role_id`` and ``secret_id`` parameters' env vars ``VAULT_ROLE_ID`` and ``VAULT_SECRET_ID`` have been replaced with ``ANSIBLE_HASHI_VAULT_ROLE_ID`` and ``ANSIBLE_HASHI_VAULT_SECRET_ID`` (https://github.com/ansible-collections/community.hashi_vault/issues/20).
+minor_changes:
+  - hashi_vault - ``url`` can be specified via new ``ANSIBLE_HASHI_VAULT_ADDR`` env var (https://github.com/ansible-collections/community.hashi_vault/issues/8).
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -42,14 +42,16 @@ DOCUMENTATION = """
     token_path:
       description: If no token is specified, will try to read the token file from this path.
       env:
-        - name: VAULT_TOKEN_PATH
+        - name: ANSIBLE_HASHI_VAULT_TOKEN_PATH
+          version_added: '1.0.0'
       ini:
         - section: lookup_hashi_vault
           key: token_path
     token_file:
       description: If no token is specified, will try to read the token from this file in C(token_path).
       env:
-        - name: VAULT_TOKEN_FILE
+        - name: ANSIBLE_HASHI_VAULT_TOKEN_FILE
+          version_added: '1.0.0'
       ini:
         - section: lookup_hashi_vault
           key: token_file
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -72,14 +72,16 @@ DOCUMENTATION = """
     role_id:
       description: Vault Role ID. Used in approle and aws_iam_login auth methods.
       env:
-        - name: VAULT_ROLE_ID
+        - name: ANSIBLE_HASHI_VAULT_ROLE_ID
+          version_added: '1.0.0'
       ini:
         - section: lookup_hashi_vault
           key: role_id
     secret_id:
       description: Secret ID to be used for Vault AppRole authentication.
       env:
-        - name: VAULT_SECRET_ID
+        - name: ANSIBLE_HASHI_VAULT_SECRET_ID
+          version_added: '1.0.0'
     auth_method:
       description:
         - Authentication method to be used.
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -82,7 +82,8 @@ DOCUMENTATION = """
       description:
         - Authentication method to be used.
       env:
-        - name: VAULT_AUTH_METHOD
+        - name: ANSIBLE_HASHI_VAULT_AUTH_METHOD
+          version_added: '1.0.0'
       ini:
         - section: lookup_hashi_vault
           key: auth_method
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -55,9 +55,10 @@ DOCUMENTATION = """
           key: token_file
       default: '.vault-token'
     url:
-      description: URL to the Vault service.
+      description: URL to the Vault service. If not set elsewhere, the value of the C(VAULT_ADDR) env var will be used.
       env:
-        - name: VAULT_ADDR
+        - name: ANSIBLE_HASHI_VAULT_ADDR
+          version_added: '1.0.0'
       ini:
         - section: lookup_hashi_vault
           key: url
@@ -318,6 +319,7 @@ LOW_PRECEDENCE_ENV_VAR_OPTIONS = {
     'token_path': ['HOME'],
     'namespace': ['VAULT_NAMESPACE'],
     'token': ['VAULT_TOKEN'],
+    'url': ['VAULT_ADDR'],
 }
 
 


changelogs/fragments/34-env-var-changes.yml
```diff
@@ -0,0 +1,8 @@
+---
+breaking_changes:
+  - hashi_vault - ``url`` parameter's env var ``VAULT_ADDR`` now has lower precedence than INI value (https://github.com/ansible-collections/community.hashi_vault/issues/8).
+  - hashi_vault - ``auth_method`` parameter's env var ``VAULT_AUTH_METHOD`` has been replaced by ``ANSIBLE_HASHI_VAULT_AUTH_METHOD`` (https://github.com/ansible-collections/community.hashi_vault/issues/17).
+  - hashi_vault - ``token_file`` and ``token_path`` parameters' env vars ``VAULT_TOKEN_FILE`` and ``VAULT_TOKEN_PATH`` have been replaced by ``ANSIBLE_HASHI_VAULT_TOKEN_FILE`` and ``ANSIBLE_HASHI_VAULT_TOKEN_PATH`` (https://github.com/ansible-collections/community.hashi_vault/issues/15).
+  - hashi_vault - ``role_id`` and ``secret_id`` parameters' env vars ``VAULT_ROLE_ID`` and ``VAULT_SECRET_ID`` have been replaced with ``ANSIBLE_HASHI_VAULT_ROLE_ID`` and ``ANSIBLE_HASHI_VAULT_SECRET_ID`` (https://github.com/ansible-collections/community.hashi_vault/issues/20).
+minor_changes:
+  - hashi_vault - ``url`` can be specified via new ``ANSIBLE_HASHI_VAULT_ADDR`` env var (https://github.com/ansible-collections/community.hashi_vault/issues/8).
```
briantist, line 8: @felixfontein opinions on formatting here? I could have done all removals in breaking, all additions in minor; or everything in breaking; could have added separate entries for the "paired" ones (`token_file` and `token_path`), etc.

For me, no clear winner among them, I figure you must have a preferred way for listing these :)

felixfontein: Definitely using both sections is the best way. I think the way you split this up is good!

plugins/lookup/hashi_vault.py
```diff
@@ -42,22 +42,25 @@
     token_path:
       description: If no token is specified, will try to read the token file from this path.
       env:
-        - name: VAULT_TOKEN_PATH
+        - name: ANSIBLE_HASHI_VAULT_TOKEN_PATH
+          version_added: '1.0.0'
```
felixfontein, line 46: (Maybe I already asked that before, but...) Is there a reason for not keeping the old names (and deprecating them)? That way, not every use of the plugin has to be checked and changed immediately, but people have a version where they get deprecation warnings and have time to remove usage of all "wrong" variables, until they are effectively removed.

changelogs/fragments/34-env-var-changes.yml
briantist: Should I then change the others so that for example, `ANSIBLE_HASHI_VAULT_AUTH_METHOD` being added goes in `minor_changes`, and the removal is described in `breaking_changes`?

plugins/lookup/hashi_vault.py
briantist: My thinking was this is  a new major version, so might as well go for removal. 
But for these 3 that are being removed, I'd be open to deprecation. Is it supported with an additional attribute? Like I can I have 2 env vars in the list where one of them is deprecated by adding a `deprecated:` field or something? I kinda remember there being something like that (a lot of chatter about how to support it in collections, whether it should be version or time-based, etc.) but I can't seem to find the docs.

briantist: Do you think it's too fast to add the new options in 0.2.0 and remove the old in 1.0.0? It's short in calendar time but it does allow for a version where both work.

changelogs/fragments/34-env-var-changes.yml
felixfontein: Yes. It's probably good to have two entries, like `The xxx option has been depreacted and will be removed in 2.0.0, use yyy instead` in `deprecated_features`, and `The yyy option has been added` in `minor_changes`. `breaking_changes` is only needed for actually breaking changes, not for deprecations.

plugins/lookup/hashi_vault.py
felixfontein: For end-users (who don't explicitly use the new collection before there will be a redirect in community.general 2.0.0), it will look like it worked in Ansible 2.10.x, and suddenly stopped working in Ansible 2.11. It should at least be deprecated for some time. There should be a proper deprecation period for Ansible users.

briantist:
Closing in favor #35 after discussion ; thank you Felix!

PR #43, opened 2020-12-27
CI - more caching: community.crypto and some vault caching tweaks

briantist:
##### SUMMARY

- Tweak the vault caching and cache key
- Download `community.crypto` from GitHub instead of Galaxy (https://github.com/ansible/galaxy/issues/2302)
- Cache the latest version as well so we're not checking it out 50+ times per run

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -144,9 +144,9 @@ jobs:
               basename $(curl -fs -o/dev/null -w %{redirect_url} $1/releases/latest)
           }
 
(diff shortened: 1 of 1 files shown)


briantist:
Well I got it working but I'm not really sure it's worth the trouble.  Might rip it out and just do simple retries from galaxy.
Might keep finding latest version and checkout but not cache. 
Might keep finding latest version and then use `ansible-galaxy` to install from git repo instead of galaxy.
Will mull it over.

Now that ansible-2.9 install time... I'd really like to bring that down; it takes about 30s... 🤔

PR #50, opened 2021-01-21
hashi_vault - add 'proxy' option

tchernomax:
##### SUMMARY

Add ``proxy`` option so it's possible access vault service that are behind a proxy without using this proxy on the entire playbook. 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

hashi_vault.py

##### ADDITIONAL INFORMATION

This **doesn't** work:
```
---
- hosts: localhost
  gather_facts: false
  tasks:
    - debug:
        msg: "{{ lookup('hashi_vault', 'secret=… url=…') }}"
      environment:
        HTTPS_PROXY: socks5://127.0.0.1:1080
```

lookups use the environments variables applied to ansible, so I need to use:
```
HTTPS_PROXY="socks5://127.0.0.1:1080" ansible-playbook toto.yml
```

But then my whole play use this proxy, which is not what I want.
This MR fix the problem by adding a proxy option:

```
---
- hosts: localhost
  gather_facts: false
  tasks:
    - debug:
        msg: "{{ lookup('hashi_vault', 'secret=… url=… proxy=socks5://127.0.0.1:1080') }}"
```


briantist:
Hi @tchernomax , thanks for your submission! 

I plan to do a more thorough review as soon as I get some time, but for early feedback I'd really like to see tests for this. Do you think you could take a shot at implementing testing for connecting via proxy?

tchernomax:
Hello @briantist 

> I'd really like to see tests for this. Do you think you could take a shot at implementing testing for connecting via proxy?

Yes I can… but for that I need a proxy, do you have a idea of how to build that in the CI ?

briantist:
> > I'd really like to see tests for this. Do you think you could take a shot at implementing testing for connecting via proxy?
> 
> Yes I can… but for that I need a proxy, do you have a idea of how to build that in the CI ?

Not yet! [Here's an example I found of testing a proxy in Ansible tests](https://github.com/ansible/ansible/blob/a077bca5d5866603cdc4a2a13aef8f416860aacb/test/integration/targets/yum/tasks/proxy.yml), it looks to be specific to `yum` so probably not directly usable, but may be a good inspiration.

Small but important distinction: we'd want the proxy tests working not just in the central CI, but via direct `ansible-test` runs, so like that example above they will be primarily written as ansible. I mention this because it'd probably be much easier to do it in CI only, however we'd like to keep compatibility with `ansible-test` so people can run tests locally.

There's some internal changes I'm working on that should make writing tests for something like this easier, but it's pretty early in development.

tchernomax:
@briantist done… (`apt install tinyproxy` is ugly but every other way I tried failed)

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tests.yml
```diff
@@ -1,76 +1,75 @@
...
         VAULT_ADDR: 'http://localhost:8200'
 
 - when: cryptography_version.stdout is version('1.6', '>=')
   block:
-    - name: 'test {{ auth_type }} auth with certs (validation enabled, lookup parameters)'
+    - name: 'test {{ auth_type }} auth{{ ", with proxy" if conn_param_proxy }}, with certs (validation enabled, lookup parameters)'
       include_tasks: '{{ auth_type }}_test.yml'
       vars:
-        conn_params: 'url=https://localhost:8201 ca_cert={{ local_temp_dir }}/cert.pem validate_certs=True '
+        conn_params: '{{ conn_param_proxy }} url=https://localhost:8201 ca_cert={{ local_temp_dir }}/cert.pem validate_certs=True '
```
briantist, line 20: Following on from the comment in `main.yml`, as I look over all these connection tests, at first glance it seems like we'd want to do proxy and no proxy tests for all of these variations, and that may be true in the future, but I need to re-examine all the invocation in here that are using environment variables. As far as I know, those shouldn't be working, so I suspect the env vars are being ignored and we're not testing what we think we are.

In any case, I think for the purposes of testing proxy for now, we should just add 2 invocations here: one using an HTTPS target and one with an HTTP target. So that would mean duplicating this highlighted test, and the first one in the file, to make a proxy variant of each.

At that point, you'd be modifying `conn_params` directly on each of those calls, so we can do away with the `conn_param_proxy` var.

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tinyproxy_server.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Install tinyproxy
+  shell: |
+    apt-get update
+    DEBIAN_FRONTEND=noninteractive apt-get --assume-yes install tinyproxy
+  changed_when: false
```
briantist, line 6: This isn't going to be portable. I recommend looking at [the `package` module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html), as this should pick the right package provider for the given OS.

The default testing containers use Ubuntu, and we use those in the CI, so it MUST work in Ubuntu, but where possible, when someone runs `ansible-test` locally, either against their local env or with a different container, we should strive for it to work.

In the case where we try to make it portable, but know for certain of environments where it can't, let's say we know it doesn't work on MacOS for example, then we need to gate the entire class of tests so that on those platforms, a test run can still succeed, even if skips some tests that it can't execute (there are some conditionals like that already within the tests for some RedHat versions, etc.). If you tried that and hit errors link me to a test run.

There is also an issue with using the `package` or `apt` modules on Ubuntu/Debian, when running in a version of python that doesn't match the "system" version of python (which _will happen_ in CI). Check `vault_download.yml` to see how I got around that.

We'll also want to add and then notify a handler that will uninstall this package at the end of the tests (not always running in a container).

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tinyproxy_server.yml
```diff
@@ -0,0 +1,19 @@
...
+
+- name: Configure tinyproxy
+  copy:
+    content: |
+      Port 8001
+      MaxClients 100
+      StartServers 10
+    dest: /tinyproxy.conf
```
briantist, line 14:
```suggestion
    dest: /tmp/tinyproxy.conf
```
Let's try to account for not running in a disposable/non-persisted FS

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tinyproxy_server.yml
```diff
@@ -0,0 +1,19 @@
...
+    apt-get update
+    DEBIAN_FRONTEND=noninteractive apt-get --assume-yes install tinyproxy
+  changed_when: false
+
+- name: Configure tinyproxy
+  copy:
+    content: |
+      Port 8001
+      MaxClients 100
+      StartServers 10
```
briantist, line 13:
```suggestion
    content: |
      Port 8001
      MaxClients 100
      StartServers 10
      PidFile /tmp/tinyproxy.pid
```

We should capture the PID somewhere so that a handler at the end of the run can kill the proxy (again keeping n mind this may not always be running in a container).

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tinyproxy_server.yml
```diff
@@ -0,0 +1,19 @@
...
+    dest: /tinyproxy.conf
+
+- name: Start tinyproxy
+  shell: |
+    tinyproxy -c /tinyproxy.conf
+  changed_when: false
```
briantist, line 19:
```suggestion
  notify: "name of cleanup handler"
```
We don't need to mess with the changed status, but as mentioned above a handler should be added that will kill the proxy at the end of the tests. These tests are configured to run handlers even when there's a failure.

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/token_test.yml
```diff
@@ -94,10 +94,19 @@
       register: test_inexistent
       ignore_errors: true
 
+    - name: 'Failure expected when wrong proxy is used'
+      vars:
+        secret_wrong_proxy: "{{ lookup('community.hashi_vault.hashi_vault', (conn_params|regex_replace('proxy=\\S*')) ~ 'proxy=http://127.0.0.1:4567 secret=' ~ vault_kv2_path ~ '/secret1 auth_method=token token=' ~ user_token) }}"
```
briantist, line 99: nit
```suggestion
        secret_wrong_proxy: "{{ lookup('community.hashi_vault.hashi_vault', (conn_params|regex_replace('proxy=\\S*', 'proxy=http://127.0.0.1:4567')) ~ 'secret=' ~ vault_kv2_path ~ '/secret1 auth_method=token token=' ~ user_token) }}"
```

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/token_test.yml
```diff
@@ -94,10 +94,19 @@
...
+      ignore_errors: true
+
     - name: 'Check expected failures'
       assert:
         msg: "an expected failure didn't occur"
         that:
           - test_wrong_cred is failed
           - test_unauthorized is failed
           - test_inexistent is failed
+          - test_wrong_proxy is failed
```
briantist, line 112: Some of these other ones could probably use similar treatment (another day, another PR), but it's better if we can confirm the failure is for the reason we expected (instead of for an unrelated reason). Feel free to tweak the text being searched for if there's something better. Test output: https://github.com/ansible-collections/community.hashi_vault/runs/1755466090#step:12:672

```suggestion
          - test_wrong_proxy is failed and test_wrong_proxy.msg is search('Cannot connect to proxy')
```

briantist (changes requested):
Great work @tchernomax , I appreciate you taking the time! Most of the inline suggestions are around the testing environment and ensuring we don't always assume a container or a specific distro.

Also want to move the selection of tests into `tests.yml` with the other connections params, but I think it will be simpler (see inline comments).

Some comments around adding handlers for the tinyproxy teardown, you can have a look at `handlers/mian.yml` for examples of definitions.

Please add an example or two in the `EXAMPLES` section in the documentation as well.

plugins/lookup/hashi_vault.py
```diff
@@ -448,6 +453,12 @@ def __init__(self, **kwargs):
         if self.options['auth_method'] == 'token':
             client_args['token'] = self.options.get('token')
 
+        if self.options.get('proxy'):
+            client_args['proxies'] = {
+                'http': self.options['proxy'],
+                'https': self.options['proxy']
+            }
```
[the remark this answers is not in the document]
tchernomax: @briantist I do not see why it's best to override only one scheme. Worse, it may fail when there is a redirection:

* url is in http://
* we use the proxy only for http
* a call to url return a redirection http→https
* the proxy isn't used → fail

Why do you want to only override one scheme ? Is it just a matter of taste… which I can understand :) or is there a argument I didn't think of.

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tests.yml
```diff
@@ -1,76 +1,75 @@
-- name: 'test {{ auth_type }} auth without SSL (lookup parameters)'
+- name: 'test {{ auth_type }} auth{{ ", with proxy" if conn_param_proxy }}, without SSL (lookup parameters)'
```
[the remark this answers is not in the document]
tchernomax: I will remove it.

# note
the 'else' part is actually optional: https://jinja.palletsprojects.com/en/2.11.x/templates/#if-expression

```
max@mde-claranet % cat toto.yml
---
- hosts: localhost
  gather_facts: false
  tasks:
    - debug:
        msg: "{{ 'toto' if False }}"
max@mde-claranet % ansible-playbook toto.yml
PLAY [localhost] *************************************************************************************************

TASK [debug] *****************************************************************************************************
ok: [localhost] => {
    "msg": ""
}

PLAY RECAP *******************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0```

**Comment by tchernomax** [comment]
@briantist 

> Also want to move the selection of tests into `tests.yml` with the other connections params, but I think it will be simpler (see inline comments).

done

> Some comments around adding handlers for the tinyproxy teardown, you can have a look at `handlers/main.yml` for examples of definitions.

done

> Please add an example or two in the `EXAMPLES` section in the documentation as well.

done (I add to shrink the other parameters else I pass the threshold of 160 column)

File: changelogs/fragments/50-add-proxy-option.yml [hunk 10]
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - hashi_vault - add ``proxy`` option
```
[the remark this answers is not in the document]
tchernomax: done

plugins/lookup/hashi_vault.py
```diff
@@ -95,6 +95,11 @@
       description: Authentication user name.
     password:
       description: Authentication password.
+    proxy:
+      description:
+        - URL to the proxy (eg. 'socks5://user:pass@host:port') to access the Vault service.
+        - If not specified, standard environment variables are used.
+      type: str
```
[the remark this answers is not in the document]
tchernomax: done,
I also change socks5 to http as socks5 need some additional dependency (which I describe in the EXEMPLE).

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tests.yml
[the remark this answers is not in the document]
tchernomax: done

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/main.yml
```diff
@@ -33,14 +54,34 @@
...
+    apply:
+      vars:
+        auth_type: jwt
+        conn_param_proxy: ''
+        default_path: "{{ auth_methods.jwt.default_path }}"
+        this_path: "{{ item }}"
+  when: ansible_distribution != 'RedHat' or ansible_distribution_major_version is version('7', '>')
+  loop: "{{ auth_methods.jwt.paths }}"
+
+- name: jwt tests with proxy
```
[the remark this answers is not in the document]
tchernomax: done

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tests.yml
tchernomax: done

> In any case, I think for the purposes of testing proxy for now, we should just add 2 invocations here: one using an HTTPS target and one with an HTTP target. So that would mean duplicating this highlighted test, and the first one in the file, to make a proxy variant of each.

I agree

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tinyproxy_server.yml
tchernomax: done

In fact on my first try I used "package" but as you describe it failed. I didn't saw you already have a workaround.

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tinyproxy_server.yml
tchernomax: done
But I used `local_temp_dir` instead of /tmp

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tinyproxy_server.yml
tchernomax: indeed

done

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tinyproxy_server.yml
tchernomax: done

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/token_test.yml
tchernomax: your suggestion seems find. I wouldn't have wrote it differently.

done

plugins/lookup/hashi_vault.py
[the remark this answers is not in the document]
briantist: > Why do you want to only override one scheme ?

The main reason was just to set only what is needed, and not overstep.

The redirection scenario is a good example, though by the same logic that scenario could fail if we set both, as it's not guaranteed that the same proxy address should be used for both schemes.

Probably the most "correct" option is to rename this param `proxies` and take in the dictionary so that it can be fully user-defined. This would work, but it would no longer be possible to pass it through the term string, nor the single env var or ini entry. It would also be optimizing for the unusual case rather than the usual, which is why I didn't suggest it earlier.

Another option is to offer both params.. `proxy` being a single string that can be used to set one or both schemes, `proxies` being available as a true dict. Yet another option, is to independently define `proxy_http` and `proxy_https`... more options further complicates testing...

Something I'm trying to consider is that while this is a single lookup plugin now, I'm looking toward centralizing all of these connection options so that they can become shared by more plugins and even modules (the latter of which can have dict params passed in much more easily), and it's important that these things work consistently.

In the end, if we go with your suggestion we should document the behavior and note that it cannot be used with independent http/https proxies; as far as I know there'd be no fallback other than defining it via the env vars and setting it for the whole run (the thing you are trying to fix in this PR).

Let me think on this a bit more, in the meantime let me know if you have additional thoughts about the above.

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tests.yml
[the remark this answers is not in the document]
briantist: Thanks, I thought I got burned by that once before, I may have been getting it confused with pure python, or may have just been confused in general 😅

briantist:
Great, I will look deeply through these updates a little later.

> > Please add an example or two in the `EXAMPLES` section in the documentation as well.
> 
> done (I add to shrink the other parameters else I pass the threshold of 160 column)

Yup no problem, I had to reduce nearly all of the examples in the past, twice. I had a big PR that started before the split to collections, but didn't merge until after, so I had to include reducing all the existing examples to be able to fit the Fully-qualified collection name. Then I had to do it again when we moved the plugin to this collection, since its name is a little bit longer!

plugins/lookup/hashi_vault.py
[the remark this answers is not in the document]
tchernomax: I think defining `proxy_http` and `proxy_https` is the simplest and most effective solution. I thought about it when I first wrote this patch, but I didn't need it for my particular use case so I went for `proxy`.
I will make the changes.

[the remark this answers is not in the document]
briantist: would recommend holding off on the changes, two options means additional tests, etc. and requires duplication of the value if  you need it for both (which isn't the worst). Going to give it some more thought.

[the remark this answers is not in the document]
tchernomax: done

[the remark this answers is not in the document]
tchernomax: I am resolving this conversation, but if this solution doesn't suit you feel free to comment.

briantist (changes requested):
Thanks for your patience @tchernomax . 

I've done some testing and I think the following is a good way forward:
* a single option named `proxies`
* uses `type: raw`
* accepts either a `str` or `dict`
* when provided with a `dict` it is sent directly to `hvac` unmodified
* when provided with a `str`, we do your original implementation: use it as a proxy for both `http` and `https`

This should give us the best of both worlds: ease of use for the most common cases, and full customization available for those who need it.

The dict can't be set in `env:` or `ini:` supplied values but I think that's ok.

What this requires to implement:
* description in the option explaining how to specify it (as I described above, but maybe more user friendly)
* entries in the examples section showing various ways to use it (I can help with that as well)
* validation of the option (ensuring either a string or dict was specified, setting the correct value, etc.). For this part, check out the  `process_options` method; it just calls some other specific purpose methods. I'd like to see a new specific method for this, and then have it called from there.
* it's a bit much to require tests to set up different proxies just to test this; we're not writing the proxy functionality, but since we are already testing both http and https, maybe we can switch it up with varying methods of sending in the proxy, so we at least ensure it works via specifying as string or dict (and maybe an "expected failure" test where we explicitly set the wrong one with dict format?).

What do you think?

tchernomax:
>     * a single option named `proxies`
>     * uses `type: raw`
>     * accepts either a `str` or `dict`
>     * when provided with a `dict` it is sent directly to `hvac` unmodified
>     * when provided with a `str`, we do your original implementation: use it as a proxy for both `http` and `https`
> 
> This should give us the best of both worlds: ease of use for the most common cases, and full customization available for those who need it.
> 
> The dict can't be set in `env:` or `ini:` supplied values but I think that's ok.

In fact it can.
I use check_type_dict so if the string is interpretable as a dict, it will be (see the EXAMPLES).

> What this requires to implement:
> 
>     * description in the option explaining how to specify it (as I described above, but maybe more user friendly)

done

>     * entries in the examples section showing various ways to use it (I can help with that as well)

done

>     * validation of the option (ensuring either a string or dict was specified, setting the correct value, etc.). For this part, check out the  `process_options` method; it just calls some other specific purpose methods. I'd like to see a new specific method for this, and then have it called from there.

done

>     * it's a bit much to require tests to set up different proxies just to test this; we're not writing the proxy functionality, but since we are already testing both http and https, maybe we can switch it up with varying methods of sending in the proxy, so we at least ensure it works via specifying as string or dict (and maybe an "expected failure" test where we explicitly set the wrong one with dict format?).

done

> What do you think?

good for me

plugins/lookup/hashi_vault.py
```diff
@@ -95,6 +95,21 @@
...
     password:
       description: Authentication password.
+    proxies:
+      description:
+        - URL(s) to the proxies used to access the Vault service.
+        - It can be a string or a dict.
+        - If it's a string (eg. 'proxies=http://user:pass@host:port') it is used to access HTTP and HTTPS resources.
+        - "If it's a dict (eg. 'proxies={\\"http\\": \\"http://host1\\", \\"https\\": \\"https://host2\\"}')"
+        - you can specify a different proxy for HTTP and HTTPS resources.
+        - If not specified, L(environment variables from the Requests library,https://requests.readthedocs.io/en/master/user/advanced/#proxies) are used.
```
briantist, line 105:
```suggestion
        - URL(s) to the proxies used to access the Vault service.
        - It can be a string or a dict.
        - If it's a dict, provide the scheme (eg. C(http) or C(https)) as the key, and the URL as the value.
        - If it's a string, provide a single URL that will be used as the proxy for both C(http) and C(https) schemes.
        - A string that can be interpreted as a dictionary will be converted to one (see examples).
        - You can specify a different proxy for HTTP and HTTPS resources.
        - If not specified, L(environment variables from the Requests library,https://requests.readthedocs.io/en/master/user/advanced/#proxies) are used.
```
It's already a long description so wanted to keep the examples to the examples section. Let me know what you think.

plugins/lookup/hashi_vault.py
```diff
@@ -323,6 +338,28 @@
...
+  ansible.builtin.debug:
+    msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret=... token=... url=https://... proxies=https://user:pass@myproxy:8080') }}"
+
+- name: 'use a socks proxy (need some additional dependencies, see: https://requests.readthedocs.io/en/master/user/advanced/#socks )'
+  ansible.builtin.debug:
+    msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret=... token=... url=https://... proxies=socks5://myproxy:1080') }}"
+
+- name: use proxies with a dict (as param)
+  ansible.builtin.debug:
+    msg: "{{ lookup('community.hashi_vault.hashi_vault', '...', proxies={'http': 'http://myproxy1', 'https': 'http://myproxy2'}) }}"
```
briantist, line 354:
```suggestion
- name: use proxies with a dict (as param)
  ansible.builtin.debug:
    msg: "{{ lookup('community.hashi_vault.hashi_vault', '...', proxies={'http': 'http://myproxy1', 'https': 'http://myproxy2'}) }}"

- name: use proxies with a dict (as param, pre-defined var)
  vars:
    prox:
      http: http://myproxy1
      https: https://myproxy2
  ansible.builtin.debug:
    msg: "{{ lookup('community.hashi_vault.hashi_vault', '...', proxies=prox }}"
```

These are all great! Could we add one additional?

plugins/lookup/hashi_vault.py
```diff
@@ -323,6 +338,28 @@
...
+
+- name: use proxies with a dict (in the term string, ansible syntax)
+  ansible.builtin.debug:
+    msg: "{{ lookup('community.hashi_vault.hashi_vault', '... proxies=http=http://myproxy1,https=http://myproxy2') }}"
+
+- name: use proxies with a dict (in the term string, plain syntax)
```
briantist, line 360:
```suggestion
- name: use proxies with a dict (in the term string, JSON syntax)
```

plugins/lookup/hashi_vault.py
```diff
@@ -323,6 +338,28 @@
...
+
+- name: 'use a socks proxy (need some additional dependencies, see: https://requests.readthedocs.io/en/master/user/advanced/#socks )'
+  ansible.builtin.debug:
+    msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret=... token=... url=https://... proxies=socks5://myproxy:1080') }}"
+
+- name: use proxies with a dict (as param)
+  ansible.builtin.debug:
+    msg: "{{ lookup('community.hashi_vault.hashi_vault', '...', proxies={'http': 'http://myproxy1', 'https': 'http://myproxy2'}) }}"
+
+- name: use proxies with a dict (in the term string, ansible syntax)
```
briantist, line 356:
```suggestion
- name: use proxies with a dict (in the term string, Ansible key=value syntax)
```

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tests.yml
```diff
@@ -73,4 +83,3 @@
           environment:
             VAULT_ADDR: 'https://localhost:8201'
             VAULT_SKIP_VERIFY: False
-
```
briantist, line 76: nit but, please preserve newline at EOF

briantist (changes requested):
This is great work @tchernomax , I love the ability to accept a serialized dict as a string, which makes for good compatibility with term string, env vars, and ini. 

I had thought about that when I proposed the dual idea but discarded it because I didn't want us to implement/support something non-standard, but I didn't know about the validation module_util so that's perfect.

I made a couple tweaks to the param description and examples, but not married to them, lmk if you don't agree.
(and just a newline that went missing)

plugins/lookup/hashi_vault.py
tchernomax: LGTM, commited

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/tests.yml
tchernomax: done

tchernomax: removing this line is necessary to pass sanity test:
https://github.com/ansible-collections/community.hashi_vault/pull/50/checks?check_run_id=1825111960

briantist: I see, that may have been my mistake, I suppose it was originally 2 blank lines (shown in diff view as 1 blank line + EOF), and I misinterpreted it. Apologies!

briantist (approved):
Looks great @tchernomax , thanks again for taking the time to contribute and working through the changes!

PR #55, opened 2021-02-08
first steps at centralizing code and adding unit tests

briantist:
##### SUMMARY
The first steps to allowing code re-use within the collection. The goal is to break things out into pieces to allow for easier, faster, and more complete testing, and then to pave the way for new plugins and modules that will all build on these shared components.

Major goals:
- clear separation between auth (login) and authenticated operations: this is how Vault really works, and ensuring the internal components have those boundaries allows us to surface the boundaries where it makes sense (plugins dedicated to auth, returning tokens), and plug them when in modularly in the internals for all-in-one plugins (like `hashi_vault`)
- where appropriate, move tests to units for faster execution and reduced integration test bloat
- where appropriate, run integration tests against the shared components, which should themselves be easier tests; further reducing the number of tests needed on individual plugins (we're past 500 on the `hashi_vault` lookup when you multiplex all the params)
- centralized auth method definition and validation (the bulk of the existing plugin), in concert with doc fragments

**In this PR**
- module util `HashiVaultHelper` containing code to be shared by plugins and (future) modules
- plugin util `HashiVaultPlugin` (inheriting from `AnsiblePlugin`), containing stuff specific to AnsiblePlugins in the collection
- `HashiVaultLookupBase` (inheriting from `LookupBase`), containing stuff specific to Lookup Plugins in the collection (I expect I'd do the same for other plugin types)
- First set of (very basic) unit tests for the new stuff above, and the existing Lookup
- Run units in CI (~broken until I can figure out how to skip python 2.6~ worked around 2.6 but unsure if it's the best way going forward, as it's unlikely 2.6 will be supported at any time because `hvac` doesn't support it)
- Moved a few small things from the lookup to the shared classes above
  - Deprecation workaround is moved to `HashiVaultPlugin`
  - Creation of hvac client and the `.logout()` action moved to `HashiVaultHelper`
  - Parsing lookup term strings is moved to `HashiVaultLookupBase`

This PR only scratches the surface, and is intended to be a start, I don't want to move too many things in one PR.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

hashi_vault


plugins/lookup/__init__.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2021 Brian Scholer (@briantist)
+# Simplified BSD License (see licenses/simplified_bsd.txt or https://opensource.org/licenses/BSD-2-Clause)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+from ansible.errors import AnsibleError
+# from ansible.plugins import AnsiblePlugin
```
felixfontein, line 9: It's probably a good idea to remove the out-commented imports before merging this, as these tend to stay around and confuse for a long time ;-)

plugins/lookup/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+from ansible.utils.display import Display
+
+from ansible_collections.community.hashi_vault.plugins.plugin_utils.hashi_vault_plugin import HashiVaultPlugin
+
+display = Display()
+
+
+class HashiVaultLookupBase(HashiVaultPlugin, LookupBase):
+
+    def parse_kev_term(self, term, first_unqualified=None, plugin_name=None):
```
felixfontein, line 21: Why not force `plugin_name` to always be specified?

plugins/plugin_utils/hashi_vault_plugin.py
```diff
@@ -0,0 +1,55 @@
...
+display = Display()
+
+
+class HashiVaultPlugin(AnsiblePlugin):
+    def __init__(self, loader=None, templar=None, **kwargs):
+        super(HashiVaultPlugin, self).__init__()
+
+        self.helper = HashiVaultHelper()
+
+    def deprecate(self, collection_name='community.hashi_vault'):
```
felixfontein, line 24: How about calling this `process_deprecations` instead? Then you don't need the pylint ignore when calling this :)

tests/unit/requirements.txt
```diff
@@ -0,0 +1 @@
+hvac ; python_version > '2.6'
```
felixfontein, line 1: Shouldn't this be:
```suggestion
hvac ; python_version >= '2.7'
```
Or does it support Python 2.6.1? (Maybe the syntax excludes 2.6.1 as well, I don't know the syntax well enough to determine that, but it's definitely clearer to everyone if you use `>=` :) )

plugins/lookup/__init__.py
briantist: agreed, right now they are reminders for me as I work on it :)

plugins/plugin_utils/hashi_vault_plugin.py
briantist: I will give it a try. I already tried adding `_` and `__` and pylint still hit it, I'll see if a more different name fixes that. It's weird because it was called `deprecate` when it was part of `hashi_vault` and yet never caused a lint problem.

plugins/lookup/__init__.py
briantist: Hm, I thought I envisioned some reason why, can't remember now, but you're probably right. Will make that change and if I actually ran into something, could reconsider.

briantist:
I think this is at a good place to get merged. No user-facing functionality changes, introduction of units so slight increase in coverage), and the beginnings of what I hope to be a more sustainable layout. Thanks for the review @felixfontein !

Next going to focus on starting to move option processing and validation centrally.

briantist:
Triggering one more test run before merging just in case, there have been some upstream test changes.
https://github.com/ansible-collections/community.hashi_vault/actions/runs/562390752

Issue #35, opened 2020-12-23
[internals] hashi_vault - plugin should always retrieve options with get_option()


briantist:
##### SUMMARY

As part of the work done in https://github.com/ansible-collections/community.general/pull/23 toward separating vault-related functionality from plugin-specific functionality, a new class was introduced to contain the vault-specific stuff.

The options from the plugin class were passed wholesale into the new class, to be used directly as a dictionary.

Although the plugin class is using `self.set_options()`, which seems sufficient to do things like process `env:` and `ini:` blocks, it may not always provide all of the functionality that that `self.get_option()` does. For example, one of the proposed fixes for https://github.com/ansible/ansible/issues/73051 involves moving the deprecation handling specifically into `get_option()`.

Fixing the issue wholesale would be a bit involved, as the most direct fix either is to pass the entire `LookupPlugin` object into the `HashiVault` class, and have `HashiVault` call `get_option()` all the time, or to increase the scope of the `validate_` methods to ensure they enumerate all optional options.

But ultimately it's a design flaw; the purpose of the `HashiVault` class was supposed to be toward moving functionality into a module_util, and having to pass the plugin object into there seems like an anti-pattern (the plugin should process relevant options and pass them into methods, but as written the plugin is only partially aware of which options are relevant for a given auth method), so I don't want to dedicate significant effort toward that type of temporary fix.

Rather, I'll approach this with a more manual two-step approach:

1. Immediate workaround: deprecation is the only possible place where this has an impact in the short term. As we add deprecations, we will add throwaway calls to `self.get_option()` in the appropriate `validate_` method for parameters that aren't already accessed (optional ones mainly), and to the lookup class's `run` method itself for params that aren't specific to an auth method (like `url`) for any option which has a deprecation, to ensure the message gets processed. Everything else stays the same.

2. As efforts toward making a module_util continue, the interface between plugin and util must and will ensure that this pattern doesn't propagate into the future, by carefully considering where responsibility lies. The interface problem will then be resolved naturally.

If other issues arise from the lack of universal `get_option()` calls before the module_util or other work can refactor to fix the core issue, then those will be dealt with at that time.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

hashi_vault.py

##### ANSIBLE VERSION

```paste below

```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```paste below

```

PR #59, opened 2021-02-22
(internals) moving more functionality centrally, more tests

briantist:
##### SUMMARY

This follows on the heels of #55 , and is a little big but I'll try to break down the changes and what's happening.

### Options Adapter
The `HashiVaultOptionAdapter` class provides a consistent interface for accessing options, and lives in module_utils.

The goal is to have logic for dealing with various common options in module_utils so that it can be shared among plugins and eventually modules as well. But options are accessed differently between plugins via Config Manager and modules via AnsibleModule.

Config Manager uses a dict internally, but it's not meant to be interacted with directly. As learned from the deprecations incident and described in #35 , the ensure the best future compatibility it's best to ensure all option access goes through Config Manager's public methods, and that options aren't accessed when not needed.

To that end I want to ensure that the logic of processing options doesn't need to know where the option came from, but passing in all the options it might need at once isn't ideal either, as it puts too much responsibility on the caller, and may cause unnecessary option access.

The adapter allows these methods to work consistently regardless of the source.

The adapter has class methods to easily create instances from a dict directly or from a AnsiblePlugin. I'll make one for AnsibleModule later (which is probably just going to call the dict version anyway).

The adapter class has a full suite of unit tests (issues caught in writing those led to a few changes in design, love it!)

---

### Option Group Base Class
`HashiVaultOptiongroupBase` is a base class for what I'm tentatively calling "option groups" and the classes that will handle them. For now the functionality of processing low preference environment variables was moved here (along with the hardcoded list of them). That's probably going to change a bit as time goes on.

### Connection Options Group
`HashiVaultConnectionOptions` is the first options group class, encompassing the options needed to make a connection to Vault. So the methods responsible for processing those options moved as well, and were changed to use the adapter.

---

### Doc fragment
The connection options were moved to a doc fragment.

There's a few other small changes, test improvements, etc. scattered in here.

---

### What's Next in this Project
In the short term: I'd introduce a diff immediately after this one that renames the module_util with an underscore `_` adds a docstring for python compatibility, and a comment specifying that it's private. This is all came from https://github.com/ansible/community/issues/539#issuecomment-780839686 . It's not in this diff because I started this before that meeting, and the rename is going to show the file as all new, so I want to separate it from other changes to not mess up the diff.

Other than that:
- I'll be looking at starting to move auth methods to dedicated classes
- I want to think about how to organize doc fragments in a way that will work better for being shared with both plugins and modules
- I want to see if some information can be extracted from doc fragments so it's not repeated elsewhere

##### COMPONENT NAME

hashi_vault

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,6 @@
 ---
+ignore:
+  - "**/tests/unit/compat/*"
+
 fixes:
   - "ansible_collections/community/hashi_vault/::"
--- a/plugins/doc_fragments/connection.py
+++ b/plugins/doc_fragments/connection.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2012, Brian Scholer (@briantist)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+class ModuleDocFragment(object):
+
+    DOCUMENTATION = r'''
+    options:
+      url:
+        description:
+          - URL to the Vault service.
+          - If not specified by any other means, the value of the C(VAULT_ADDR) environment variable will be used.
+          - If C(VAULT_ADDR) is also not defined then a default of C(http://127.0.0.1:8200) will be used.
+        env:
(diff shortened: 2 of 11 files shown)


felixfontein (approved):
Great work!

PR #71, opened 2021-03-22
hashi_vault - add retries option to allow for retrying failed requests

### LINKED ISSUE #58: hashi_vault - add options for retries on failure
##### SUMMARY
Vault being an HTTP service, there are numerous reasons why a single request could fail to complete, and it would be great to have a way to retry those requests.

Unlike tasks, there's no easy way to "retry" a lookup plugin (or indeed most controller-side plugin types) on the whole, so this isn't easily solved outside the plugin itself.

Disorganized list of thoughts/considerations:
- offer some sort of backoff, maybe several backoff strategies
- include jitter
- be smart about which failures to retry on (not access denied, not found, etc.), can it be configurable without being too complicated?
- [`backoff` module](https://pypi.org/project/backoff/) would provide dead-simple implementation, but an additional dependency
- can [`requests`/`urllib3` retries](https://findwork.dev/blog/advanced-usage-python-requests-timeouts-retries-hooks/#retry-on-failure) be used? 
  - is that an add'l dependency, or does it come with `requests` already? 
  - Could that be used properly with [`hvac` adapters](https://hvac.readthedocs.io/en/stable/advanced_usage.html#custom-requests-http-adapter)? If so, may be best supported solution, though more work to implement.

This will probably not be implemented until things move a little more central, so that it can be tested more easily, independently of specific plugins.

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION

```yaml

```

briantist:
##### SUMMARY

Closes #58 

Adds `retries` option which enables retrying failed requests with `urllib3.Retry`.

Also adds `retry_action` to control whether or not to emit warnings when a request is retried.

This PR also adds the start of a collection docsite, with a User Guide that goes into detail on the retry support.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

hashi_vault

##### ADDITIONAL INFORMATION

```paste below

```


docs/extra-docs.yml
```diff
@@ -0,0 +1,4 @@
+---
```
felixfontein, line 1: This needs to be in `docs/docsite/`, not in `docs/`. (The `docsite/` is also missing in the next file's path.)

plugins/doc_fragments/connection.py
```diff
@@ -99,4 +99,41 @@ class ModuleDocFragment(object):
           - name: ansible_hashi_vault_timeout
         type: int
         version_added: '1.3.0'
+      retries:
+        description:
+          - "Allows for retrying on errors, based on
+            the L(Retry class in the urllib3 library,https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry)."
+          - This collection defines recommended defaults for working with Vault.
+          - This option can be specified as a boolean, number, or dictionary.
+          - If this option is not specified or has a value of C(False) or C(0), then retries are disabled.
```
felixfontein, line 108:
```suggestion
          - If this option is not specified or has a value of C(false) or C(0), then retries are disabled.
```
Ansible documentation should use lower-case strings for YAML booleans.

felixfontein (commented):
You probably want to run `pip install antsibull && antsibull-lint collection-docs .` in a new workflow to add basic linting for the docs.

(I only looked at the documentation.)

plugins/doc_fragments/connection.py
briantist: @felixfontein where is that defined? In YAML (including in Ansible) I usually use title case, but would like to be aware of what's recommended.

bcoca: This is not an Ansible requirement but a docs style issue

plugins/module_utils/_hashi_vault_common.py
```diff
@@ -214,26 +240,143 @@ def process_low_preference_env_vars(self, option_vars=None):
...
+        retries_opt = self._options.get_option('retries')
+
+        if retries_opt is None:
+            return
+
+        # we'll start with a copy of our defaults
+        retries = self._RETRIES_DEFAULT_PARAMS.copy()
+
+        try:
+            # check for floats first, because check_type_int won't accept them, but check_type_bool will
```
felixfontein, line 353: `check_type_bool` isn't used anymore.

briantist:
It looks I've gotten about all the review I'm gonna get on this one. In addition to tests added here, I've run other local manual tests. I learned some things with that, that I hope to port over to CI at some point, but it's going to require some more of the test improvement work to be done. It's pretty exciting nonetheless, looking at using MMock to support some advanced testing scenarios that are going to be difficult with a real Vault server alone (intermittent failing connections, maybe AWS/Azure/GCP/etc. backend testing without setting up the actual backend 3rd party service, etc.) by mocking various endpoints.

In the meantime, I'll merge this is in so we can get it released.

PR #73, opened 2021-04-06
run CI against Ansible 2.12/stable-2.11

briantist:
##### SUMMARY

Re: https://github.com/ansible-collections/overview/issues/45#issuecomment-813821100

Ensure we test against 2.11-stable and the new devel (2.12)

##### ADDITIONAL INFORMATION

```paste below

```


briantist:
although python 3.5 is no longer supported for ansible-2.12 on the controller, the tests passed, so I'm leaving it for now. If that changes as breaking changes are introduced into devel, I'll exclude those.

PR #80, opened 2021-05-15
hashi_vault - add support for `none` auth type

### LINKED ISSUE #79: Feature Request : Support Vault Agent Credentials
##### SUMMARY
Support the use of a locally running Vault Agent process to interact with Vault.  This relieves the user almost entirely have need to manage or interact with tokens.  Example case:

* Vault Agent is configured to perform AWS IAM authentication.
* The Vault Agent has a TCP listener on 127.0.0.1:8100 (standard HashiCorp deployment examples)
* Ansible using `hashi_vault` then just instructs `hvac` to connect to the local listener and authentication is transparent.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`community.hashi_vault`, authentication methods.

##### ADDITIONAL INFORMATION

Related: https://github.com/ansible/ansible/issues/60728
It appears that this was suggested before, but deemed unsupported by the upstream Python `hvac` module, but the notes in the docs there suggest that the module is missing support for *UNIX Sockets* connections to the Vault Agent.  This works fine when connecting to a local TCP Vault Agent listener.

jlrgraham23:
##### SUMMARY
Support the use of a locally running Vault Agent process to interact with Vault. This relieves the user almost entirely have need to manage or interact with tokens.

Fixes #79 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
community.hashi_vault, authentication methods.

##### ADDITIONAL INFORMATION

More details in: https://github.com/ansible-collections/community.hashi_vault/issues/79

```
- name: authenticate with vault agent
  ansible.builtin.debug:
    msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/hello:value', auth_method=agent, url='http://127.0.0.1:8100') }}"
```

Also related, older issue against the main Ansible repo: https://github.com/ansible/ansible/issues/60728

### DIFF
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -147,7 +147,7 @@ DOCUMENTATION = """
         - approle
         - aws_iam_login
         - jwt
-        - agent
+        - none
       default: token
     return_format:
       description:
@@ -308,14 +308,14 @@ EXAMPLES = """
   ansible.builtin.debug:
     msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/hello:value', token=my_token, token_validate=False) }}"
 
-# Agent auth
+# None auth
 
-# Vault Agent running with a TCP listener on port 8100.
+# For use with Vault Agent running with a TCP listener on port 8100.  Where the agent will handle authentication to Vault.
 # https://www.vaultproject.io/docs/agent
 
 - name: authenticate with vault agent
   ansible.builtin.debug:
-    msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/hello:value', auth_method='agent', url='http://127.0.0.1:8100') }}"
+    msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/hello:value', auth_method='none', url='http://127.0.0.1:8100') }}"
 
 # Use a proxy
 
@@ -570,7 +570,7 @@ class HashiVault:
         except (NotImplementedError, AttributeError):
             raise AnsibleError("JWT authentication requires HVAC version 0.10.5 or higher.")
 
-    def auth_agent(self):
+    def auth_none(self):
         pass
     # end auth implementation methods
 
@@ -642,7 +642,7 @@ class LookupModule(HashiVaultLookupBase):
     def auth_methods(self):
         # enforce and set the list of available auth methods
         # TODO: can this be read from the choices: field in documentation?
-        avail_auth_methods = ['token', 'approle', 'userpass', 'ldap', 'aws_iam_login', 'jwt', 'agent']
+        avail_auth_methods = ['token', 'approle', 'userpass', 'ldap', 'aws_iam_login', 'jwt', 'none']
         self.set_option('avail_auth_methods', avail_auth_methods)
         auth_method = self.get_option('auth_method')
 
@@ -737,6 +737,6 @@ class LookupModule(HashiVaultLookupBase):
     def validate_auth_jwt(self, auth_method):
         self.validate_by_required_fields(auth_method, 'role_id', 'jwt')
 
-    def validate_auth_agent(self, auth_method):
+    def validate_auth_none(self, auth_method):
         pass
     # end auth method validators
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -155,6 +155,7 @@ DOCUMENTATION = """
         - C(dict) returns a single dict containing the key/value pairs.
         - C(values) returns a list of all the values only. Use when you don't care about the keys.
         - C(raw) returns the actual API result (deserialized), which includes metadata and may have the data nested in other keys.
+        - C(none) auth method was added in collection version C(1.2.0).
       choices:
         - dict
         - values
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -315,7 +315,7 @@ EXAMPLES = """
 
 - name: authenticate with vault agent
   ansible.builtin.debug:
-    msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/hello:value', auth_method=agent, url='http://127.0.0.1:8100') }}"
+    msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/hello:value', auth_method='agent', url='http://127.0.0.1:8100') }}"
 
 # Use a proxy
 
@@ -512,7 +512,6 @@ class HashiVault:
     #    0.9.6 -- userpass
     #    0.10.5 -- jwt (new)
     #    0.10.6 -- approle
-    #    1.2.0 -- agent
     #
     def auth_token(self):
         if self.options['auth_method'] == 'token':
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -147,6 +147,7 @@ DOCUMENTATION = """
         - approle
         - aws_iam_login
         - jwt
+        - agent
       default: token
     return_format:
       description:
@@ -306,6 +307,15 @@ EXAMPLES = """
 - name: authenticate without token validation
   ansible.builtin.debug:
     msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/hello:value', token=my_token, token_validate=False) }}"
+
+# Agent auth
+
+# Vault Agent running with a TCP listener on port 8100.
+# https://www.vaultproject.io/docs/agent
+
+- name: authenticate with vault agent
+  ansible.builtin.debug:
+    msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/hello:value', auth_method=agent, url='http://127.0.0.1:8100') }}"
 
 # Use a proxy
 
@@ -560,6 +570,8 @@ class HashiVault:
         except (NotImplementedError, AttributeError):
             raise AnsibleError("JWT authentication requires HVAC version 0.10.5 or higher.")
 
+    def auth_agent(self):
+        pass
     # end auth implementation methods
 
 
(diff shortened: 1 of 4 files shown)


plugins/lookup/hashi_vault.py
```diff
@@ -132,6 +132,7 @@
         - approle
         - aws_iam_login
         - jwt
+        - agent
```
Akasurde, line 135: Add a note in the `description` stating when this is supported from.

jlrgraham23: Thanks @Akasurde , could you be more specific?  I don't see any particular `description` fields that seem to have specific notes about the versions when an auth method is added.  I added a comment that seemed to match the versions of the added methods.  Let me know if this isn't what you were wanting.

briantist (changes requested):
So first off, I want to thank you for the submission and apologize for the delay in reviewing. I've been on vacation (still am) so I may be slower to respond than usual.

As for the general idea, I have a few thoughts.

First, I want to confirm that agent auth can't already be achieved, by using "token" auth and passing anything for the token value, like
```
lookup('community.hashi_vault.hashi_vault', 'secret/hello:value', auth_method='token', token='fake')
```

I played around with agent a little in the past and I thought this worked. I ask not because that means we don't need or want something more appropriate, but because it's good to know workarounds/alternatives.

As for your addition, the method you're adding, called "agent" is not actually doing anything specifically, it's just passing nothing. So I'm wondering if this should be handled differently, like maybe having an "auth" method called `none` instead? After all it is also possible to use an agent with other auth methods, so calling this "agent" could be confusing imo.

The other major issue here is that there are no tests. I'm working hard to refactor the plugin, and to improve testing, and hopefully to make it easier to add tests for future contributions, but I'm hesitant to add new features that don't include any tests.

Related to naming, if we're calling the auth method "agent" then we're sort of saying it's expected to be used with agent, and that means we should test against agent. Since we don't yet test against a Vault agent it'll be some doing to get that going in CI.

Testing against Vault itself, and against other software (like `tinyproxy` which we use to test proxy operations), can be kind of pain, and I'm working on making that experience better right now.

On the other hand, if we call this auth method `none`, I think we can do with much less testing of it; basically we'd probably just against Vault as normal but we'd expect an access denied in our tests.

`none` will also flow better if there are other ways of running Vault server/agent in the future that make use of no authentication.

What do you think?

plugins/lookup/hashi_vault.py
```diff
@@ -471,6 +481,7 @@ def get(self):
     #    0.9.6 -- userpass
     #    0.10.5 -- jwt (new)
     #    0.10.6 -- approle
+    #    1.2.0 -- agent
```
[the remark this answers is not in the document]
jlrgraham23: Hi @briantist!  Thanks for the great feedback.  Upfront, if you're on vacation no need at all to respond to this now!  I've got a local build and distribution system in my environment where this change is needed, so it's not blocking usage.  Just wanted to try and contribute back.

I like the idea of renaming the method to `none`, I'd just pulled `agent` out as it seemed to be more descriptive of what was actually happening -- that is, the user is declaring that they want to delegate the authentication, but at the same time, I guess that isn't actually really that different from just saying default to the token in `${HOME}/.vault-token`.  If that simplifies testing then great, I'll update for that convention.

I'm also happy to try and help with getting any CI tests in for Agent, but I'm new to this testing framework.  I'll try to dig at some docs and see if I can figure out what's needed! :)

Some notes on if this is needed, unfortunately just passing in a fake token errors out when the plugin tries to validate the token.
```
lookup("community.hashi_vault.hashi_vault", "secret= secret/hello:value", auth_method="token", token="fake")
```
Yields:
```
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.hashi_vault.hashi_vault'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Invalid Hashicorp Vault Token Specified for hashi_vault lookup."}
```

And explicitly disabling validation 
```
lookup("community.hashi_vault.hashi_vault", "secret=secret/hello:value", auth_method="token", token="fake", token_validate=false)
```
Yields:
```
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.hashi_vault.hashi_vault'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Forbidden: Permission Denied to secret 'secret/hello'."}```
```

I did a quick dig on the Vault Agent docs and didn't find an explicit answer (it was a cursory dig), but it appears that the behavior if a token is supplied it overrides Vault Agent's credentials.

Without `VAULT_TOKEN` set or with a `${HOME}/.vault-token` file and the agent running
```
$ echo $VAULT_ADDR
http://127.0.0.1:8100
$ /usr/local/bin/vault kv get secret/hello
====== Metadata ======
Key              Value
---              -----
created_time     2021-03-29T03:38:36.993280771Z
deletion_time    n/a
destroyed        false
version          1

====== Data ======
Key         Value
---         -----
value       foo
```

With an invalid token:
```
$ VAULT_TOKEN=foo /usr/local/bin/vault kv get secret/hello
Error making API request.

URL: GET http://127.0.0.1:8100/v1/sys/internal/ui/mounts/secret/hello
Code: 403. Errors:

* permission denied
```

With an different valid token:
```
$ VAULT_TOKEN=<itsasecret> /usr/local/bin/vault kv get secret/hello
====== Metadata ======
Key              Value
---              -----
created_time     2021-03-29T03:38:36.993280771Z
deletion_time    n/a
destroyed        false
version          1

====== Data ======
Key         Value
---         -----
value       foo
```

jlrgraham23:
Hi @briantist!  Thanks for the great feedback.  Upfront, if you're on vacation no need at all to respond to this now!  I've got a local build and distribution system in my environment where this change is needed, so it's not blocking usage.  Just wanted to try and contribute back.

I like the idea of renaming the method to `none`, I'd just pulled `agent` out as it seemed to be more descriptive of what was actually happening -- that is, the user is declaring that they want to delegate the authentication, but at the same time, I guess that isn't actually really that different from just saying default to the token in `${HOME}/.vault-token`.  If that simplifies testing then great, I'll update for that convention.

I'm also happy to try and help with getting any CI tests in for Agent, but I'm new to this testing framework.  I'll try to dig at some docs and see if I can figure out what's needed! :)

Some notes on if this is needed, unfortunately just passing in a fake token errors out when the plugin tries to validate the token.
```
lookup("community.hashi_vault.hashi_vault", "secret= secret/hello:value", auth_method="token", token="fake")
```
Yields:
```
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.hashi_vault.hashi_vault'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Invalid Hashicorp Vault Token Specified for hashi_vault lookup."}
```

And explicitly disabling validation 
```
lookup("community.hashi_vault.hashi_vault", "secret=secret/hello:value", auth_method="token", token="fake", token_validate=false)
```
Yields:
```
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.hashi_vault.hashi_vault'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Forbidden: Permission Denied to secret 'secret/hello'."}```
```

I did a quick dig on the Vault Agent docs and didn't find an explicit answer (it was a cursory dig), but it appears that the behavior if a token is supplied it overrides Vault Agent's credentials.

Without `VAULT_TOKEN` set or with a `${HOME}/.vault-token` file and the agent running
```
$ echo $VAULT_ADDR
http://127.0.0.1:8100
$ /usr/local/bin/vault kv get secret/hello
====== Metadata ======
Key              Value
---              -----
created_time     2021-03-29T03:38:36.993280771Z
deletion_time    n/a
destroyed        false
version          1

====== Data ======
Key         Value
---         -----
value       foo
```

With an invalid token:
```
$ VAULT_TOKEN=foo /usr/local/bin/vault kv get secret/hello
Error making API request.

URL: GET http://127.0.0.1:8100/v1/sys/internal/ui/mounts/secret/hello
Code: 403. Errors:

* permission denied
```

With a different valid token:
```
$ VAULT_TOKEN=<itsasecret> /usr/local/bin/vault kv get secret/hello
====== Metadata ======
Key              Value
---              -----
created_time     2021-03-29T03:38:36.993280771Z
deletion_time    n/a
destroyed        false
version          1

====== Data ======
Key         Value
---         -----
value       foo
```

plugins/lookup/hashi_vault.py
```diff
@@ -132,6 +132,7 @@
         - approle
         - aws_iam_login
         - jwt
+        - none
       default: token
     return_format:
       description:
```
briantist, line 138: @jlrgraham23 I agree with @Akasurde , a note should be added saying that the `none` method was added in version `$next`. Before the plugin was moved to a collection, which sort of reset its versioning, there were other notes like this when they were relevant. You can see such an example here: https://github.com/ansible-collections/community.hashi_vault/commit/9841595d9017755b5449b144470c078f5276c503#diff-c47dc8233f99f31fc4a283e177d047b83d3df18d8e52f25a3ca3b375bffce0e9R47

briantist (changes requested):
Changes are looking good so far, how do you feel about trying to add the tests for the `none` method based on the tests for the other auth methods? I'd probably go off the `token` auth method but the difference would mainly be that you wouldn't be supplying a token and that you'd expect the calls to fail, so it should be simpler to set up.

briantist:
Hi @jlrgraham23 I took the liberty of adding tests and a changelog fragment, and making a few other test updates, hope that's ok. Could you take a look and let me know if it looks good to you? And if possible to try it out with a Vault agent as you seem to have one set up already.

@Akasurde since I'm now a co-author would you be willing to give this another look so there's some additional review?

plugins/lookup/hashi_vault.py
```diff
@@ -147,13 +147,15 @@
...
         - Controls how multiple key/value pairs in a path are treated on return.
         - C(dict) returns a single dict containing the key/value pairs.
         - C(values) returns a list of all the values only. Use when you don't care about the keys.
         - C(raw) returns the actual API result (deserialized), which includes metadata and may have the data nested in other keys.
+        - C(none) auth method was added in collection version C(1.2.0).
```
felixfontein, line 158: This needs to go to `auth_method`. Right now this is in the description of `return_format`.

briantist: oof, good find, thank you!

felixfontein (approved):

Issue #83, opened 2021-05-22
Remove default Vault address (http://127.0.0.1:8200) for the url option and make it required


briantist:
The default value for the `url` option is `http://127.0.0.1:8200`, which is meant to somewhat correspond to the "defaults" used by Vault CLI and the `hvac` library.

For several reasons, the default is unlikely to be correct for production uses (it doesn't use TLS, runs on localhost, etc.).

Instead, my observation in real world use cases is that people tend to forget to set a Vault address and then are confused about why it isn't working because they see the access denied, timed out, etc. error message but don't really notice that they aren't connecting to the correct Vault server.

As a result, I think the default does more harm than good.

I've also received feedback along the same lines ([here's one example](https://github.com/ansible-collections/community.hashi_vault/issues/49#issuecomment-784399943)), and I'm tending to agree with the critics.

We also provide many ways of specifying the address:
- direct argument
- INI config
- multiple environment variables
- (upcoming #49) ansible vars

So anyone who is using the default for its intended value, has several other ways of not specifying it, including several that don't require modification of playbooks/tasks/templates.

**Nonetheless it would be a breaking change and would happen in the next major version**

---

## Request for Feedback

If you have any opinion on this change either way, I invite you to provide it here!

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

hashi_vault.py

briantist:
cc @Akasurde @elcomtik @pilou-

pilou-:
> Remove default Vault address (http://127.0.0.1:8200) for the url option and make it required

:+1:
Would the option considered as set when `VAULT_ADDR` environment variable is set?

briantist:
> > Remove default Vault address (http://127.0.0.1:8200) for the url option and make it required
> 
> 👍
> Would the option considered as set when `VAULT_ADDR` environment variable is set?

@pilou-

Just as now, `VAULT_ADDR` or the stronger precedence `ANSIBLE_HASHI_VAULT_ADDR` environment variables could still be used.

None of the proactive ways of setting the value would be different by way of this change; only if the value is not specified in any way, is the default used now; this change would cause that case to be an error instead.

elcomtik:
@briantist I totally agree with proposal, it reflects all my needs.

PR #86, opened 2021-05-30
Add ability to use ansible vars to specify options

### LINKED ISSUE #49: AWX and ANSIBLE_ prefixed environment variables - support a 2nd set of env vars? ansible vars?
##### SUMMARY
Nothing against your new environment variable naming scheme (https://github.com/ansible-collections/community.hashi_vault/issues/10#issue-754561671), but would you consider supporting a second set of environment variables which does not use the `ANSIBLE_` prefix but maybe something like `ANS_` or similar?

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

hashivault lookup plugin

##### ADDITIONAL INFORMATION

The problem we have with the `ANSIBLE_` prefix is that `AWX` does not allow user credentials to induce environment variables starting with this prefix.

> When creating a new credential type, you are responsible for avoiding collisions in the extra_vars, env, and file namespaces. Also, avoid environment variable or extra variable names that start with ANSIBLE_ because they are reserved.

[SOURCE](https://docs.ansible.com/ansible-tower/latest/html/userguide/credential_types.html)

### LINKED ISSUE #65: remove templating in conditionals in tests
##### SUMMARY

There's templating in conditionals in a bunch of tests. They work fine but throw warnings since conditionals are already "in" a template.

This is just a reminder to go through those and replace the templating.

##### ADDITIONAL INFORMATION

```yaml

```

briantist:
##### SUMMARY
Fixes #49 
Fixes #65

### The big change here is adding `vars:` entries for several options:
- `url ->`
  - `ansible_hashi_vault_url`
  - `ansible_hashi_vault_addr`
- `proxies -> ansible_hashi_vault_proxies`
- `namespace -> ansible_hashi_vault_namespace`
- `token -> ansible_hashi_vault_token`
- `token_validate -> ansible_hashi_vault_token_validate`
- `role_id -> ansible_hashi_vault_role_id`
- `secret_id -> ansible_hashi_vault_secret_id`
- `auth_method -> ansible_hashi_vault_auth_method`

Some new examples of usage:

```yaml
- name: use ansible vars to supply some options
  vars:
    ansible_hashi_vault_url: 'https://myvault:8282'
    ansible_hashi_vault_auth_method: token
  set_fact:
    secret1: "{{ lookup('secret/data/secret1') }}"
    secret2: "{{ lookup('secret/data/secret2') }}"

- name: use proxies with a dict (as direct ansible var)
  vars:
    ansible_hashi_vault_proxies:
      http: http://myproxy1
      https: https://myproxy2
  ansible.builtin.debug:
    msg: "{{ lookup('community.hashi_vault.hashi_vault', '...' }}"
```

## Important Note
Using templating in task vars **won't work** because of https://github.com/ansible/ansible/issues/73268 so this example **will fail**:
```yaml
- name: use ansible vars to supply some options
  vars:
    ansible_hashi_vault_url: '{{ my_other_url }}'
  debug:
    msg: "{{ lookup('secret/data/secret1') }}"
```

Instead, use `set_fact` to ensure your vars are evaluated beforehand.

```yaml
- set_fact:
    ansible_hashi_vault_url: '{{ my_other_url }}'

- name: use ansible vars to supply some options
  debug:
    msg: "{{ lookup('secret/data/secret1') }}"
```

### Other Changes
- There were many unnecessary test loops around the `validate_certs` option, checking different types of ways to specify yes/no, but they used environment variables via the `environment:` keyword which doesn't affect the lookup anyway. They have been removed, as unit tests now cover the various ways of supplying that option.
- Other test loops also erroneously used environment variables to set the Vault URL; that wasn't actually being used, it was actually using the default option value, so those would have broken after #83 . They have been replaced with ansible var equivalents.
- An additional proxy test loop was added that specified the `proxies` option as a dict via an ansible var.
- Templating in conditionals was removed to get rid of some warnings in the test output (#65).
- Removed a conditional about cryptography version that's not needed anymore.
- Other areas have been changed to use ansible vars to get more coverage on them or to simplify a test or both. I expect more replacements in the future to greatly cut down on the amount of string concatenation we're doing in the term string.
- A stronger evaluation of the token validation with no default policy expected failure test, by adding an assert to ensure the failure happens and with the message we expect.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

hashi_vault

### DIFF
--- a/changelogs/fragments/86-add-vars-options.yml
+++ b/changelogs/fragments/86-add-vars-options.yml
@@ -1,0 +1,10 @@
+---
+minor_changes:
+  - hashi_vault lookup - add ``ansible_hashi_vault_url`` and ``ansible_hashi_vault_addr`` Ansible vars entries to the ``url`` option (https://github.com/ansible-collections/community.hashi_vault/pull/86).
+  - hashi_vault lookup - add ``ansible_hashi_vault_proxies`` Ansible vars entry to the ``proxies`` option (https://github.com/ansible-collections/community.hashi_vault/pull/86).
+  - hashi_vault lookup - add ``ansible_hashi_vault_namespace`` Ansible vars entry to the ``namespace`` option (https://github.com/ansible-collections/community.hashi_vault/pull/86).
+  - hashi_vault lookup - add ``ansible_hashi_vault_token`` Ansible vars entry to the ``proxies`` option (https://github.com/ansible-collections/community.hashi_vault/pull/86).
+  - hashi_vault lookup - add ``ansible_hashi_vault_token_validate`` Ansible vars entry to the ``proxies`` option (https://github.com/ansible-collections/community.hashi_vault/pull/86).
+  - hashi_vault lookup - add ``ansible_hashi_vault_role_id`` Ansible vars entry to the ``proxies`` option (https://github.com/ansible-collections/community.hashi_vault/pull/86).
+  - hashi_vault lookup - add ``ansible_hashi_vault_secret_id`` Ansible vars entry to the ``proxies`` option (https://github.com/ansible-collections/community.hashi_vault/pull/86).
+  - hashi_vault lookup - add ``ansible_hashi_vault_auth_method`` Ansible vars entry to the ``proxies`` option (https://github.com/ansible-collections/community.hashi_vault/pull/86).
--- a/plugins/doc_fragments/connection.py
+++ b/plugins/doc_fragments/connection.py
@@ -22,6 +22,11 @@ class ModuleDocFragment(object):
         ini:
           - section: lookup_hashi_vault
             key: url
+        vars:
+          - name: ansible_hashi_vault_url
+            version_added: '1.2.0'
+          - name: ansible_hashi_vault_addr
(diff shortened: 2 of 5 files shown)


felixfontein (approved):

briantist:
cc @h3m5k @erinn @morco @elcomtik @pilou- @wenottingham 

Would you be able to try out this PR with AWX and see if it meets your needs for supplying values that don't rely on environment variables?

h3m5k:
I'v tested this with AWX 15.0.1 and ansible 2.10.6 and it works like a charm.
The vars included in the test were (on separate occasions):
ansible_hashi_vault_url
ansible_hashi_vault_addr
ansible_hashi_vault_token
ansible_hashi_vault_role_id
ansible_hashi_vault_secret_id
ansible_hashi_vault_auth_method

Let me know if you'd like me to try something else.
Thanks a bunch!

briantist:
@h3m5k great! thanks a lot, I really appreciate that extra testing. Given how important it is to AWX users I wanted someone to use it hands-on before merging in case anything else was needed.

Issue #99, opened 2021-06-21
hashi_vault connection - add option to control timeout


briantist:
##### SUMMARY

The timeout is currently fixed to `hvac`'s default of 30s with no way to customize it.

When everything is working correctly it's unlikely it'll ever be hit. When a URL is bad or the server is down, 30s can feel like a lifetime. As I implement retry support (#71), I realize that this point compounds a lot if the request is hanging waiting for a timeout on every retry as well.

When the request legitimately needs to take longer (some auth methods or secret engines could presumably take time), it would be impossible for it to ever finish.

Note on the above: I have seen a case of AWS auth for example being hit so often that it triggers throttling errors on the AWS APIs, and Vault's internal retries (which are not configurable) have it keep it going for longer than 30s, and in this case the timeout for this plugin proved problematic, as it was our main way of triggering widespread AWS auth calls and we couldn't get the "real" response from the Vault server.

---

There should be a configurable `timeout` option to better control this.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

hashi_vault

PR #100, opened 2021-06-21
hashi_vault lookup - add `timeout` connection option

briantist:
##### SUMMARY

Fixes: #99

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

hashi_vault

### DIFF
--- a/tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/defaults/main.yml
+++ b/tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/defaults/main.yml
@@ -111,3 +111,7 @@ vault_key_file: '{{ local_temp_dir }}/privatekey.pem'
 
 vault_proxy_server: 'http://127.0.0.1:8001'
 vault_proxy_external: False
+
+# if any connections are taking longer than this to complete there's probably something really wrong
+# with the integration tests, so it'd be better to fail faster than the 30s default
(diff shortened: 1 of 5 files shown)


tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/defaults/main.yml
```diff
@@ -111,3 +111,7 @@ vault_key_file: '{{ local_temp_dir }}/privatekey.pem'
 
 vault_proxy_server: 'http://127.0.0.1:8001'
 vault_proxy_external: False
+
+# if any connections are taking longer than this to complete there's probably something really wrong
+# with the integration tests, so it'd be better to fail faster than the 30s default
+ansible_hashi_vault_timeout: 5
```
jinoa-fh, line 117: Very nice - I wonder if calling that in the description would be beneficial. Although you would reference a value from another project that is a moving target.

felixfontein (approved):

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/defaults/main.yml
briantist: I had it in there, when I had set the plugin's default for the value to 30s, but I thought I'd rather it "fall through" instead of having its own default, I just didn't update that comment.

PR #106, opened 2021-07-11
tests - massive test refactor

### LINKED ISSUE #19: hashi_vault: tests fail (unstable) on macOS 10.15
##### SUMMARY
The tests for the `hashi_vault` lookup plugin fail on macOS (added to CI in ansible-collections/community.general#971). They are currently disabled.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
plugins/lookup/hashi_vault.py

##### ANSIBLE VERSION
2.10

briantist:
##### SUMMARY

This is a big refactor of the integration tests that I've been trying to do in smaller bits and pieces for many months. Some of the more difficult and entangled aspects were not possible or practical to do in smaller increments, and I failed several times trying in previous PRs. So as much as I prefer avoid such giant changes, this was an area I found difficult to make meaningful progress otherwise.

### Changes
- The tests no longer run as `runme.sh` tests, and now use the "standard" role-based mechanism.
- The parts of the tests responsible for setting up the proxy server and the Vault server have been completely decoupled from the lookup tests and split into their own `setup_` roles.
- The integration tests now expect that addresses for Vault and a proxy server and any future external dependencies will be passed in via vars, so setting these things `integration_config.yml` is now required.
- The "legacy" all-in-one type of setup where the tests download, extract, and run a Vault server (and install, configure, run a proxy server), is still available for now, but it must be requested specifically (opt-in).
- A sample `integration_config.yml.sample` file is included. It can be copied directly to `integration_config.yml` to go right back to the old style all-in-one setup. This will prove to be less and less viable for running all tests as more targets get added, because that process will be repeated for every target, so it's best for one-time/infrequent contributors who will run only a single target locally.
- A concept of a "localenv" setup role is introduced, which are basically roles/playbooks that are intended to setup the dependencies outside of the test environment, and generate an integration_config that you can use to then run the tests. That gets you some re-use of those things, simplifying and speeding up local test runs.
- Included is a "docker" localenv that uses docker-compose to run the Vault server and proxy server in containers. I expect this to be the most used/most popular option.
- The GitHub workflows have been updated to use a "gha" localenv setup, which is just a special case of the docker one that uses pre-generated certificates and skips the `community.docker` tasks (using the already included `docker-compose` CLI), in order to remove a dependency on `community.crypto` and `community.docker` in the "main" CI test runs, and avoid the time taken to run those tasks.
- New test runs have been added to test the "legacy" modes of operation (since those aren't used in main CI anymore), and to test the docker localenv, as well as running `ansible-test` with `--docker` and with `--venv`. These tests run on `ubuntu` and `macos`, to try to ensure that we're providing a good experience for contributors.
- The addition of the above tests led to several fixes of the tests to get them running correctly, especially on MacOS.
- Getting docker installed on MacOS runners is a PITA; I combined the set of shell steps into a GitHub action.
- The local tests do still have external collection dependencies, now depending on `community.crypto`, `community.docker`, and `community.general` (macos only). Since installing from Galaxy can still be error prone, I've chosen to continue installing directly from GitHub, but I've wrapped this up into a neat GitHub Action as well, so it's quite clean in the workflow.
- New guides have been added to the docsite, around contributions and local testing. Those will continue to be expanded over time.

### What's missing?
There's a few things that are NOT done yet in this PR

- The actual configuration of a running Vault server (setting up secrets, enabling and configuring backends) is still in the lookup target. Moving that out will be the next step, but I want to get this localenv stuff and the documentation out because it's already useful as is (and this PR is already too big).
- No additional stuff that could be moved to its own target (like connection tests) has been moved yet. This will probably be the next thing as soon as the Vault runtime config mentioned above is moved out. Once that's done I think we're well set to continue moving functionality out of the lookup plugin and into module_utils, and subsequently separating out integration testing for it. And that will lead straight into being able to make new plugins and modules.

Fixes #19 (that issue is barely relevant anymore, but the tests that now run on MacOS did surface a few issues that are now fixed)

##### ADDITIONAL INFORMATION

```paste below

```


docs/docsite/rst/contributor_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Running legacy mode tests in a controlled python virtual environment (**not recommended**):
+
+.. code-block:: bash
+
+    $ ansible-test integration --venv --requirements --allow-destructive -v
+
+Note that your system packages may be manipulated by running locally or in a venv.
+
+Legacy mode is not recommended because a new Vault server and proxy server will be downloaded, set up, configured, and/or uninstalled, for every *target*. Traditionally, we've only had one target, and so it was a good way to go, but that's no longer going to be the case. This is going to make it slower and slower as you'll incur the overhead on every target, in every run.
```
acozine, line 72: This is a really convincing argument. I'd move it up to the beginning of the section, so folks might read it before they follow the instructions on running tests in legacy mode.

docs/docsite/rst/contributor_guide.rst
```diff
@@ -0,0 +1,146 @@
...
+
+As a result, if you're going to use legacy mode it's best to limit your test run to the specific target needed, for example:
```
acozine, line 74: If you do move the paragraph above, then this line would need to change - maybe something like:
```suggestion
If you must use legacy mode testing, you can make it more efficient by limiting your test run to the specific target needed, for example:
```

docs/docsite/rst/localenv_developer_guide.rst
```diff
@@ -0,0 +1,69 @@
...
+  :depth: 2
+
+
+Required External Dependencies
+------------------------------
+
+HashiCorp Vault
+...............
+
+A Vault server is required for the integration tests. Using `Vault Dev Server Mode <https://www.vaultproject.io/docs/concepts/dev-server>`_ is recommended as it's the easiest and fastest way to get a server started.
```
acozine, line 22: I'd make this section a bulleted list of requirements, with the three absolute requirements (Vault server, root token of the Vault server, and HTTP listener) listed first, then the TLS "if you need it" requirement at the end.

acozine (commented):
This looks great, thanks for creating this documentation @briantist!

I made a few suggestions for making it even better. Incorporate the changes you like and ignore the ones you don't like.

docs/docsite/rst/contributor_guide.rst
```diff
@@ -0,0 +1,146 @@
+.. _ansible_collections.community.hashi_vault.docsite.contributor_guide:
+
+Contributor Guide
+=================
```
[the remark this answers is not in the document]
briantist: Thank you so much for this link! I've re-done the headers on all three pages to (hopefully) match. That also inspired lots of other formatting changes to make the docs fit better, and I tried to apply more of the style guide recommendations on header style (Sentence case), language style, etc.

I also added an entire new section specifically on contributing documentation, and updated all the links I could to `:ref:`erences (this took a fair bit of trial and error, the doc build workaround from Sam was essential for me to test all this out).

docs/docsite/rst/contributor_guide.rst
briantist: Perfecto 👌 moved into a nice note box, slightly re-worded, thank you!

docs/docsite/rst/localenv_developer_guide.rst
briantist: I tried out a few permutations of this but in the end I think I want to keep it as is, at least for the time being. I think the table better showcases the actual values that need to be supplied. I'll definitely keep thinking about this though. It may be easier to see rendered, I'm going to see if I can attach rendered versions in the comments as well.

briantist:
@acozine thank you so much for the review! Your comments sparked a whole set up of updates to the docs, in both formatting and content. If you have some time I'd really appreciate another look. 

I'm also attaching rendered HTML versions of each of the pages, though the `:ref:` links and other content may not work right, unless you plop them into a pre-built coresite `docs/docsite/_build/html` directory. So I'm also including a ZIP of the full docsite build which might be easier.

[just my pages (html.zip)](https://github.com/ansible-collections/community.hashi_vault/files/6833987/html.zip)
[whole Ansible docsite (full.zip)](https://github.com/ansible-collections/community.hashi_vault/files/6834001/full.zip)

acozine (commented):
Overall looks great. I let my inner copyeditor loose on the text. As always, keep the changes you like and ignore the ones you don't.

Oh, one warning - if you accept any changes to headings. you'll need to extend the underlining to match, since I didn't make suggestions on those lines.

docs/docsite/rst/contributor_guide.rst
```diff
@@ -0,0 +1,222 @@
...
+
+#. Build the docs:
+
+    .. code-block:: shell-session
+
+      $ make coredocs
+
+The rendered HTML docs should be available in ``docs/docsite/_build/html/`` and can be opened in a browser.
+
+Running Tests Locally
```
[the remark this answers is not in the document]
briantist: This is not related to documentation anymore; I thought it being a peer header to the documentation section marked it enough. It looks ok to me rendered, especially in the TOC
image

docs/docsite/rst/localenv_developer_guide.rst
```diff
@@ -0,0 +1,72 @@
...
+==============================
+
+HashiCorp Vault
+---------------
+
+A Vault server is required for the integration tests. Using `Vault Dev Server Mode <https://www.vaultproject.io/docs/concepts/dev-server>`_ is recommended as it's the easiest and fastest way to get a server started.
+
+A unencrypted (plain HTTP) listener is *required* for our purposes as most of the tests will expect to connect that way.
+
+To run the tests that deal specifically with TLS/HTTPS access, the Vault server will need to be started with a TLS enabled listener as well, and the address/port, as well as the CA cert (or the cert itself if self-signed) will need to be supplied.
```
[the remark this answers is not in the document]
briantist: Updated with slightly different wording in a separate commit.

docs/docsite/rst/contributor_guide.rst
```diff
@@ -0,0 +1,222 @@
...
+  :local:
+  :depth: 3
+
+
+Quick start
+===========
+
+#. Log into your GitHub account.
+#. Fork the `ansible-collections/community.hashi_vault repository <https://github.com/ansible-collections/community.hashi_vault>`_ by clicking the **Fork** button in the upper right corner. This will create a fork in your own account.
+#. Clone the repository locally, following :ref:`the example instructions here <hacking_collections>` (but replace ``general`` with ``hashi_vault``). **Pay special attention to the path structure.**
```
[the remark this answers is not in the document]
briantist: Updated!

docs/docsite/rst/contributor_guide.rst
```diff
@@ -0,0 +1,222 @@
...
+
+*****************
+Contributor guide
+*****************
+
+This guide aims to help PR authors contribute to the ``community.hashi_vault`` collection.
+
+.. note::
+
+	This guide is a work-in-progress and should not be considered complete. Check back often as we fill out more details based on experience and feedback, and please let us know how this guide can be improved.
```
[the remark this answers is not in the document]
briantist: Yes exactly, added that (with link)

docs/docsite/rst/user_guide.rst
```diff
@@ -56,12 +59,12 @@ Current Defaults (always check the source code to confirm the defaults in your s
 Any of the ``Retry`` class's parameters that are not specified in the collection defaults or in your custom dictionary, are initialized using the class's defaults, with one exception: the ``raise_on_status`` parameter is always set to ``false`` unless you explicitly added it your custom dictionary. The reason is that this lets our error handling look for the expected ``hvac`` exceptions, instead of the ``Retry``-specfic exceptions. It is recommended that you don't override this as it may cause unexpected error messages on common failures if they are retried.
 
 Retry messages
```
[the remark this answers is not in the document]
briantist: Tracking is not quite right, but I can't come up with anything better. To me, `Retry messages` makes sense, but it's important that it make sense to everyone else too. Does it help that this is already within the `Retrying failed requests` section? 

To help explain, this section is about the option that controls whether you see a warning message when a retry is performed (for now warn or nothing are the only options, in the future it could be something else, like logging).

docs/docsite/rst/contributor_guide.rst
```diff
@@ -0,0 +1,222 @@
+.. _ansible_collections.community.hashi_vault.docsite.contributor_guide:
+
+*****************
+Contributor guide
```
acozine, line 4: Hm, the GitHub standard is to add a `contributing.md` file, right? Maybe add one and point to this file or to the built page?

docs/docsite/rst/contributor_guide.rst
```diff
@@ -0,0 +1,222 @@
...
+Additions to the collection documentation are very welcome! We have three primary types of documentation, each with their own syntax and rules.
+
+Module and plugin documentation
+-------------------------------
+
+This type of documentation gets generated from structured YAML, inside of a Python string. It is included in the same code that it's documenting, or in a separate Python file, such as a doc fragment. Please see the :ref:`module format and documentation guidance <developing_modules_documenting>` for more information.
+
+This type of documentation is tested with ``ansible-test sanity`` and full instructions are available on the :ref:`testing module documentation <testing_module_documentation>` page.
+
+Although we can't preview how the documentation will look for these, we can be reasonably assured the output is correct because of the highly structured nature and the validation via sanity tests.
```
acozine, line 40:
```suggestion
Although we can't preview how the documentation will look for these, we can be reasonably sure the output is correct because the documentation is highly structured validated using sanity tests.
```

docs/docsite/rst/contributor_guide.rst
```diff
@@ -0,0 +1,222 @@
...
+    .. code-block:: shell-session
+
+      $ make coredocs
+
+The rendered HTML docs should be available in ``docs/docsite/_build/html/`` and can be opened in a browser.
+
+Running tests locally
+=====================
+
+If you're making anything more than very small or one-time changes, you'll want to run the tests locally to avoid having to push a commit for each thing, and waiting for the CI to run tests.
```
acozine, line 97: Sneaky future tense here:
```suggestion
If you're making anything more than very small or one-time changes, run the tests locally to avoid having to push a commit for each thing, and waiting for the CI to run tests.
```

acozine (commented):
Docs look great, thanks @briantist. I found a few last nits.

docs/docsite/rst/contributor_guide.rst
acozine: ugh, I deleted the `and` there by mistake . . . "highly structured and validated using . . . "

docs/docsite/rst/contributor_guide.rst
briantist: I will leave `contributing.md` for a later time, but I have updated the (previously blank) section in the README.

docs/docsite/rst/contributor_guide.rst
briantist: oh nice, I can update your suggestion and then commit it

docs/docsite/rst/user_guide.rst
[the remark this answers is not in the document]
briantist: resolved over IRC

PR #113, opened 2021-07-28
split auth methods into module utils

briantist:
##### SUMMARY

Fixes: #116 
Resolves: #35 

Pulling the auth out of the lookup and into their own auth classes. Getting us one step closer to new plugins, and drastically reducing the size and complexity of the lookup plugin.

## Major Changes
- new `HashiVaultAuthMethodBase` class to serve as a base for all auth methods to be in their own classes (with unit tests)
- every auth method pulled out of the `hashi_vault` lookup plugin, and put into its own auth class 🎉
- new `HashiVaultAuthenticator` class serves as a single entry point that holds instances of all the other auth classes. Any content that needs to authenticate can use this one class, and doesn't need to be directly updated when new auth methods are introduced. Units also added for this class.
- The `HashiVault` class is no longer needed and removed from the lookup plugin; everything in it has been moved out to auth methods, except the `get` function which moved back into the `LookupModule` class.
- With those changes above, all plugin options access is now done correctly through Config Manager (#35) 🎉
- Imports and other stuff not needed is removed from the lookup plugin.
- auth method docstrings moved to a doc_fragment (this will need updates later, but for now it's all together in a single fragment)
- `HashiVaultConnectionOptions` class moved to its own file.

### Minor Changes
- fixed a bad import fallback for retries (#116)
- coverage flag to file mappings added (in anticipation of rebasing once #114 lands)
- a unit test for missing retry support was added

##### ADDITIONAL INFORMATION

```paste below

```


briantist:
Well,  I've tried for a few days to get reviews but no takers. I'm pretty confident in this changes so I'll be merging! No change in functionality here, and the one bugfix hasn't affected any user that I'm aware of.

PR #114, opened 2021-07-28
tests - coverage, flags, and other testing/CI updates

briantist:
##### SUMMARY

This PR started out as a simple test to see if we could move the uploading of codecov reports to happen after all the tests complete. The idea was to avoid the situation where codecov "fails" (showing the PR as a failed CI status and sending an email) due to a reduction in coverage, because it hasn't yet received all the reports (because some tests are still running).

That part went well, by uploading the reports as "artifacts" we can re-download them in a final job within the workflow that doesn't start until the rest complete, and then do the upload from there.

But then I went and asked about the `--group-by` options and started looking at "flags" in codecov, and went all out trying to support those.

`ansible-test coverage xml` with `--group-by` can split up reports by various things, but the hard part is that the codecov GitHub action (and really the underlying uploader) can't really handle that on its own; we'd have to make several calls. And since we can't loop over steps in GHA, and we can't use other actions in composite actions, and since matrix would spawn lots of separate runners and compound the concurrency issues we already have (plus possibly reintroduce the "failed" status), my only option left was to write some code to parse the fields out of the file names and make its own calls to the codecov uploader.

This is in the form of another composite action called `ansible-codecov`. It first downloads and verifies the signature of the codecov uploader.

Then it calls a python script that can take pattern forms to generate flags based on directory and/filenames. It makes a call for each flags, and in this way we can get flags for the various fields that `ansible-test` already gives us, in addition to adding custom flags. Since I upload all the report artifacts with a naming that includes Ansible version, that is also parsed out and becomes a flag.

Right now, I have every possible `--group-by` option enabled, and have it creating All The Flags. It might be a bit much, I'm going to leave it that way and see how it works out and adjust in the future to possibly remove some or otherwise make changes. I also expect that the changes in #113 (or a related PR) will add many new targets, as the auth methods get their tests moved to dedicated targets and out of the lookup plugin's tests.

Getting the "percentage" coverages right on the target flags is going to be a WIP; we need to add each one individually to `codecov.yml` to tell it which path(s) count toward that flag, but this should work out well for most plugins, and hopefully for the centralized auth methods, as long they get put into their own individual `.py` files. EDIT: see https://github.com/ansible-collections/community.hashi_vault/pull/113/commits/f851557a78dfeeb8608ca9626ef2defb243fd7cd (this commit included in #113 )

### other small changes
- exclude devel integration tests with python < 3.8 (because we have only controller side code, for now; we'll reintroduce when modules are added)
- add `tests/config.yml` (will be helpful for devel/2.12+ but doesn't affect earlier)
- remove `-v` from units (output is too big to look through meaningfully)

## tl;dr of major changes
- upload coverage reports as artifacts, then upload them all at the end of the workflow run (avoids spurious "failures" on coverage reduction)
- add additional `--group-by` options to `ansible-test coverage xml`
- add `ansible-codecov` action to process coverage reports into flags, and do individual flag uploads


mattclay (commented):
I took a quick look at the flags usage and don't see any issues.

PR #119, opened 2021-08-02
Build docsite preview on PR and push

briantist:
##### SUMMARY

Building off of @felixfontein 's work in https://github.com/ansible-community/antsibull/pull/297

Changes here build the docsite and attach a ZIP of the resultant HTML as a build artifact. See an example run here:
https://github.com/ansible-collections/community.hashi_vault/actions/runs/1088508910

A PR author or maintainer/reviewer could download that zip and take a look at the generated docs.

I intend to open another PR that I will create directly off a branch on this repo rather than a fork (due to CI permissions), to try to get the workflow to safely comment on a PR with a link to the build/artifact.

##### ISSUE TYPE

- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


docs/preview/rst/collections/index.rst
```diff
@@ -0,0 +1,14 @@
+:orphan:
```
felixfontein, line 1: This file will get deleted by the rsync in `build.sh`.

briantist: @felixfontein I see! In CI it doesn't matter, but for anyone trying to use it locally, that wouldn't be good 😅.
Since I'm using the result of `antsibull-docs init` and not actually running it each time though, removing this means it wouldn't be available for use later right?

How do you think I should proceed then? Run init as part of the process (I guess .gitignoring all the files it creates)? Should I remove the rsync step?

briantist: Resolved based on IRC discussion. Thanks Felix!

PR #143, opened 2021-09-15
[DNM] try out split controller testing

briantist:
##### SUMMARY

DO NOT MERGE

This is running our tests using the split controller changes that are upcoming for `ansible-test` to ensure our stuff still works and that we didn't make any bad assumptions about running on the controller.

##### ISSUE TYPE

- Tests Pull Request

##### COMPONENT NAME

tests/integration

##### ADDITIONAL INFORMATION

See https://github.com/ansible/ansible/pull/75605

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -192,7 +192,7 @@ jobs:
           - 3.7
           - 3.8
           - 3.9
-          - 3.10
+          - '3.10'
         exclude:
           - ansible: stable-2.9
             python: 3.9
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -33,10 +33,12 @@ jobs:
         test_container:
           - default
         ansible:
-          - stable-2.9
-          - stable-2.10
-          - stable-2.11
-          - devel
+          # - stable-2.9
+          # - stable-2.10
+          # - stable-2.11
+          # - devel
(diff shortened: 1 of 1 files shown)


briantist:
https://github.com/ansible/ansible/pull/75605 has been merged, so this PR is no longer necessary.

As of now, CI still passes, but there are some TODOs for this collection that will be addressed in one or more other PRs:
* ensure all instances of the test modules (used in tests only) use `delegate_to: localhost` -- they actually work as is, because they are used to talk to the running Vault server, and access works in docker and in a venv, from the target, but they really should be run on the controller alongside the Vault server.
* the `auth_*`  and `connection_options` targets don't have a clear context, so there are warnings about that; they will run on the target. That's the correct place to run them (they are module_utils) but let's provide `context/target` in `aliases` to make it official (`context/controller` is also possible, in case we ever need it explicitly).
* will need to re-include additional python versions in devel CI, to properly cover supported target python versions again
* group controller vs target tests in `aliases` so that they can be invoked differently, and we can avoid invalid combinations like python versions that aren't supported on the controller, and/or specify different controller and target python versions
* the action we use to pre-download docker images is broken in devel now. CI still works because that step is allowed to fail (`ansible-test`'s invocation will do it). This will remain broken until `ansible-test` has the functionality built in, at which point we'll use the action for pre-2.12 versions of Ansible, and we'll use the built-in for devel/2.12+.

PR #144, opened 2021-09-22
common options - change ini section, deprecate old, separate plugin doc fragments

briantist:
##### SUMMARY

The currently used INI section, `lookup_hashi_vault`, is specific to the `hashi_vault` lookup, but the options have all been moved to doc fragments with the intention for them to be shared collection-wide, in both modules and plugins.

To those ends, this PR:
- adds an INI section `hashi_vault_collection` that replaces `lookup_hashi_vault`
- deprecates `lookup_hashi_vault` INI section in version 3.0.0
- ensures the `lookup_hashi_vault` INI section is only valid within the `hashi_vault` lookup, and won't be inherited by new content; unfortunately to do this, I had to duplicate all the options within that plugin, including the shared INI sections. So that's going to be crappy for maintenance for a while.
- also note if a setting is set in both the `[lookup_hashi_vault]` and `[hashi_vault_collection]` sections, the latter wins.
- splits out the plugin-specific option values in the doc fragments, preparing for module content
- updates `hashi_vault` lookup and the test lookups to use the new doc fragment sections

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

plugins/doc_fragments

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/plugins/doc_fragments/auth.py
+++ b/plugins/doc_fragments/auth.py
@@ -42,7 +42,7 @@ class ModuleDocFragment(object):
       token_validate:
         description:
           - For token auth, will perform a C(lookup-self) operation to determine the token's validity before using it.
-          - Disable if your token doesn't have the C(lookup-self) capability.
+          - Disable if your token does not have the C(lookup-self) capability.
         type: boolean
         default: true
         version_added: 0.2.0
--- a/plugins/doc_fragments/auth.py
+++ b/plugins/doc_fragments/auth.py
@@ -15,21 +15,6 @@ class ModuleDocFragment(object):
         description:
           - Authentication method to be used.
           - C(none) auth method was added in collection version C(1.2.0).
-        env:
-          - name: VAULT_AUTH_METHOD
-            deprecated:
-              why: standardizing environment variables
-              version: 2.0.0
-              collection_name: community.hashi_vault
-              alternatives: ANSIBLE_HASHI_VAULT_AUTH_METHOD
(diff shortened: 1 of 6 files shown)


plugins/doc_fragments/auth.py
```diff
@@ -49,14 +34,52 @@ class ModuleDocFragment(object):
...
           - The order of token loading (first found wins) is C(token param -> ansible var -> ANSIBLE_HASHI_VAULT_TOKEN -> VAULT_TOKEN -> token file).
+      token_path:
+        description: If no token is specified, will try to read the I(token_file) from this path.
+      token_file:
+        description: If no token is specified, will try to read the token from this file in I(token_path).
+        default: '.vault-token'
+      token_validate:
+        description:
+          - For token auth, will perform a C(lookup-self) operation to determine the token's validity before using it.
+          - Disable if your token doesn't have the C(lookup-self) capability.
```
acozine, line 45:
```suggestion
          - Disable if your token does not have the C(lookup-self) capability.
```

acozine (commented):
I didn't look at the implementation/larger implications of this change, but I did find one very small nit in the docs.

plugins/doc_fragments/auth.py
briantist: nice, thanks, that was moved not introduced in this PR, but good a time as any to improve it!

tadeboro (approved):
I see no apparent issues with this PR, but I did not test the code since I am not intimately familiar with the Hashi Vault.

Issue #138, opened 2021-09-05
docs build - process notes and TODOs


briantist:
##### SUMMARY

This issue is meant to keep track of things related to the docs build process, which builds and publishes collection documentation on PRs and pushes, in temporary sites outside of the published collection docs on docs.ansible.com.

### On a PR (open, updated, resynced)
- The docs from the target are built (rendered)
- The docs (docsite changes only) from the PR are built
- Hashes of both are recorded to determine if the PR makes any differences
- A `git diff` of the builds is taken to show in markdown
- If there are changes, then:
  - The rendered docs are attached to the build as an artifact (expires in 7 days)
  - The rendered docs are published via https://surge.sh to a temporary site that's specific to the PR. This lets anyone browse the fully rendered docs by clicking a link.
  - A bot comment on the PR provides links to the rendered sites of both the target branch docs and the PR's docs for comparison, and shows the `git diff` output in an expandable section (unless it's too big, which crashes GHA), and it shows a list of modified files, each of which links to the published docsite.
- If there are no changes:
  - if there's an existing surge site (because there were previously changes in the PR), it is unpublished
  - if there's an existing PR comment (because of previous docs changes), it is removed.

In addition to the above, another build is triggered which does exactly the same thing, _but_ it will build the entire set of docs using the entire PR contents. That means **it will include changes to plugin and module documentation**.

The key for this one is that it gets built in a separate GitHub "environment" with rules that require it to be approved on every commit. This happens because the build is still privileged and we cannot trust the PR contents, so every commit must be reviewed by a person before approving this build. But once it runs, we get the complete picture of the built docs.

### On PR close or merge
- Any existing surge site is unpublished
- If there's an existing PR comment, it is updated
  - if the PR was closed, the comment is updated to say the site is unpublished
  - if the PR was merged, the comment is updated thanking for the contribution, and linking to the main surge docsite build (as the changes have now been incorporated there)

### On Push (to `main` only)
- the docs are built, and the "main" surge site is updated.

### The `pull_request_target` event
**REQUIRED READING:** https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

When a GitHub Workflow is run for the `pull_request` event, from a fork of the repository (the norm in this and most open source projects), the rights are limited, compared to PRs opened from a branch on the same repo (which requires write access to the repo). In particular, on a PR from a fork:
[...]

PR #150, opened 2021-10-01
docs - docs-build all with approval

briantist:
##### SUMMARY

This adds a new workflow that uses a custom github "environment" so that running this workflow requires approval.
This workflow is very similar to the existing docs build, the difference is that it uses the _entire_ PR contents, and not just the `docs/docsite`.

As described in #138 , that's a security risk when running via `pull_request_target`, because the workflow has access to secrets and an elevated GitHub token.

That's where pre-approval comes in: this workflow will not run until someone has reviewed the PR and then allows this workflow to run.

The existing docsite-only workflow will still run, automatically, so PR authors can get quick feedback on those changes, and (after review) full docs rendering that includes plugin and module doc changes is still possible before a PR is merged, to the benefit of both PR authors and reviewers/maintainers.

## See it in action
A sample PR against this branch can be seen in #149 .

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

docs build

##### ADDITIONAL INFORMATION

This is not necessarily in the best end state. It is a lot of copy-paste from the other docs-build. Many of these steps should be generalized out and made into composite actions, and then they can be DRYed up a but. But I feel the advantage of getting this in place early outweighs some duplication. It will be cleaned up over time.

### DIFF
--- a/.github/workflows/docs-full.yml
+++ b/.github/workflows/docs-full.yml
@@ -1,14 +1,9 @@
 ---
-name: Collection Docs
+name: Collection Docs (Full)
 concurrency: docs-${{ github.ref }}
 on:
-  push:
-    branches:
-      - main
   pull_request_target:
-    types: [opened, synchronize, reopened, closed]
-  schedule:
-    - cron: '0 13 * * *'
+    types: [opened, synchronize, reopened]
 
 env:
   NAMESPACE: community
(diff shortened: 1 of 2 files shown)


webknjaz:
@briantist FYI this is something you may want to consider turning into a Composite Action for reusability.

briantist:
> @briantist FYI this is something you may want to consider turning into a Composite Action for reusability.

Agreed, it's already on my list! Have talked about it in IRC, and it's also in the "Additional Information" section in this PR :)

PR #157, opened 2021-10-17
[DNM] Docs/test pr/large diff

briantist:
##### SUMMARY

Docs build fails when the diff of the rendered changes is very large, because the build workflow stores the diff content in an environment variable, and when passing it to the next step in the workflow, it breaks it completely, with `Error: Argument list too long`. It's the same issue as described here: https://github.community/t/maximum-length-for-the-comment-body-in-issues-and-pr/148867

So, trying to fix the build by 1) making `git diff` possibly give better/smaller diff output and 2) if the diff output is greater than some arbitrary length, replace it with an output that only shows the files changed, rather than the content itself. That can be still be useful to be able to browse the published docs and look at specific files.

Commits here include content from another PR because that PR triggered it by having enough changes in the full docs build. This PR will NOT be merged; it's only for testing the `pull_request_target` workflows.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

docs build

##### ADDITIONAL INFORMATION

GitHub Actions has some unspecified limit to the size of arguments sent to its steps and having an environment variable that's too big can trigger this: https://github.community/t/maximum-length-for-the-comment-body-in-issues-and-pr/148867


briantist:
proper close (for site teardown)

PR #172, opened 2021-11-02
add action group

briantist:
##### SUMMARY

Adds an action group for modules in the collection. There's only 1 module right now but several others are planned.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

runtime.yml

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/changelogs/fragments/172-action_group.yml
+++ b/changelogs/fragments/172-action_group.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - community.hashi_vault collection - add the ``community.hashi_vault.vault`` action group (https://github.com/ansible-collections/community.hashi_vault/pull/172).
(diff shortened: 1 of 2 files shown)


felixfontein (approved):

dmsimard:
@briantist that's a good idea :+1:

Will you provide docs or examples how to leverage it like in https://docs.ansible.com/ansible/latest/user_guide/playbooks_module_defaults.html ?

briantist:
> Will you provide docs or examples how to leverage it like in https://docs.ansible.com/ansible/latest/user_guide/playbooks_module_defaults.html ?

@dmsimard yeah I have a lot of plans for extended documentation, including having a full section or page for many/most of the common options shared across the collection (so far there's detailed guidance for `retries:` for example in the docsite). But in trying to work against the deadlines, I will probably defer some of that.

PR #173, opened 2021-11-03
remove deprecated environment variables for 2.0.0

briantist:
##### SUMMARY

Removes deprecated env vars for the 2.0.0 release:
- Resolves: #15 
- Resolves: #17
- Resolves: #20

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

auth.py

##### ADDITIONAL INFORMATION

This completes the breaking changes that came from #10

### DIFF
--- a/changelogs/fragments/173-deprecated-env-vars.yml
+++ b/changelogs/fragments/173-deprecated-env-vars.yml
@@ -1,3 +1,3 @@
 ---
 removed_features:
-  - "community.hashi_vault collections - support for the following deprecated environment variables has been removed: ``VAULT_AUTH_METHOD``, ``VAULT_TOKEN_PATH``, ``VAULT_TOKEN_FILE``, ``VAULT_ROLE_ID``, ``VAULT_SECRET_ID`` (https://github.com/ansible-collections/community.hashi_vault/pull/173)."
+  - "support for the following deprecated environment variables has been removed: ``VAULT_AUTH_METHOD``, ``VAULT_TOKEN_PATH``, ``VAULT_TOKEN_FILE``, ``VAULT_ROLE_ID``, ``VAULT_SECRET_ID`` (https://github.com/ansible-collections/community.hashi_vault/pull/173)."
--- a/changelogs/fragments/173-deprecated-env-vars.yml
+++ b/changelogs/fragments/173-deprecated-env-vars.yml
@@ -1,0 +1,3 @@
+---
+removed_features:
+  - "community.hashi_vault collections - support for the following deprecated environment variables has been removed: ``VAULT_AUTH_METHOD``, ``VAULT_TOKEN_PATH``, ``VAULT_TOKEN_FILE``, ``VAULT_ROLE_ID``, ``VAULT_SECRET_ID`` (https://github.com/ansible-collections/community.hashi_vault/pull/173)."
--- a/plugins/doc_fragments/auth.py
+++ b/plugins/doc_fragments/auth.py
@@ -125,12 +125,6 @@ class ModuleDocFragment(object):
             version_added: 1.2.0
       token_path:
         env:
-          - name: VAULT_TOKEN_PATH
-            deprecated:
-              why: standardizing environment variables
-              version: 2.0.0
-              collection_name: community.hashi_vault
-              alternatives: ANSIBLE_HASHI_VAULT_TOKEN_PATH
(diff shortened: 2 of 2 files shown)


felixfontein (approved):

changelogs/fragments/173-deprecated-env-vars.yml
```diff
@@ -0,0 +1,3 @@
+---
+removed_features:
+  - "community.hashi_vault collections - support for the following deprecated environment variables has been removed: ``VAULT_AUTH_METHOD``, ``VAULT_TOKEN_PATH``, ``VAULT_TOKEN_FILE``, ``VAULT_ROLE_ID``, ``VAULT_SECRET_ID`` (https://github.com/ansible-collections/community.hashi_vault/pull/173)."
```
briantist, line 3:
```suggestion
  - "support for the following deprecated environment variables has been removed: ``VAULT_AUTH_METHOD``, ``VAULT_TOKEN_PATH``, ``VAULT_TOKEN_FILE``, ``VAULT_ROLE_ID``, ``VAULT_SECRET_ID`` (https://github.com/ansible-collections/community.hashi_vault/pull/173)."
```

felixfontein:
recheck

felixfontein (approved):

gdykeman:
@briantist @felixfontein 
Prepending the variable names with `ansible_*` | `ANSIBLE_*` breaks use of custom credentials in AWX / Ansible Automation Platform. Not sure of best way to resolve but wanted to bring to your attention.

briantist:
@gdykeman this has been discussed in https://github.com/ansible-collections/community.hashi_vault/issues/49

The alternative is to use Ansible vars instead which do not have the same restriction in AWX.

gdykeman:
@briantist Correct, the concern is that sensitive variables are exposed. Thus we were hoping to utilize custom credentials which do now allow for ansible to be prepended. We can stick with modifying the docfragments if it creates too much of a problem.

felixfontein:
I'm not sure why AWX/Tower doesn't like these, but prefixing with `ANSIBLE_` for env vars is what modules and plugins should do. See https://github.com/ansible/ansible/issues/52967 and the Ansible project meeting where this was [discussed and decided](https://meetbot.fedoraproject.org/ansible-meeting/2019-02-26/ansible_core_irc_public_meeting.2019-02-26-19.02.log.html).

briantist:
> @briantist Correct, the concern is that sensitive variables are exposed. Thus we were hoping to utilize custom credentials which do now allow for ansible to be prepended. We can stick with modifying the docfragments if it creates too much of a problem.

AWS credentials can inject its values as either environment variables or as Ansible vars. By injecting as Ansible vars, you will avoid this issue, because the `ansible_` prefix restriction only applies to environment variables, not to Ansible vars.

The issue I linked is where others brought up the concerns, and using Ansible vars completely solves the issue, there should be no reason to modify any docs fragments.

briantist:
> I'm not sure why AWX/Tower doesn't like these, but prefixing with `ANSIBLE_` for env vars is what modules and plugins should do. See [ansible/ansible#52967](https://github.com/ansible/ansible/issues/52967) and the Ansible project meeting where this was [discussed and decided](https://meetbot.fedoraproject.org/ansible-meeting/2019-02-26/ansible_core_irc_public_meeting.2019-02-26-19.02.log.html).

I agree with you and have tried to argue the same. Here's a partial answer: https://github.com/ansible/awx/issues/10275#issuecomment-849122944

I guess the issue is that you could use those creds to widely control Ansible behavior in ways beyond plugin configuration? But yeah, it does mean that you cannot use env vars in that way to assign options, so plugins should need to be specific about accepting `vars:` in addition (which I think is good practice anyway).

gdykeman:
This is the check:
https://github.com/ansible/awx/blob/devel/awx/main/fields.py#L816

Appreciate the quick responses!

briantist:
> This is the check: https://github.com/ansible/awx/blob/devel/awx/main/fields.py#L816
> 
> Appreciate the quick responses!

Yes, again that is the check for **environment variables**, if you choose to inject as **Ansible vars** instead, there is so such check. Please read the linked thread.

briantist:
@gdykeman see:
- https://github.com/ansible-collections/community.hashi_vault/issues/49#issuecomment-788774213

gdykeman:
@briantist @felixfontein 
This clears it up. Thank you for looking into this. Will stick to Ansible vars!

Issue #81, opened 2021-05-20
Dropping Python 2 & 3.5 Support


briantist:
# UPDATE: Python 3.5 support will also be dropped

As clarified in the `hvac` thread, support for Python 3.5 will also be dropped along with 2.x, and for the same reasons below, this collection won't be able to easily support 3.5.

**Python 3.6 will be the minimum version for `community.hashi_vault` starting in version `2.0.0`, keeping parity with `hvac`.**

---

# Python 2 Support Will be Dropped in v2
As of the next major version,  v2 (v2.0.0+) release of `community.hashi_vault`, Python 2 will no longer be directly supported or tested against. At that time the minimum Python version will likely be Python 3.5, or 3.6 (tbd).

## Why?
We all know Py2 is EoL and projects are dropping support for it left and right, including Ansible (on the controller, where currently all content in this collection runs). This makes it harder and harder to keep supporting it, to work around CI and testing issues too. **But the major driver for this decision is around the `hvac` library.**

All of the content in this collection currently relies on `hvac` and is expected to indefinitely.

`hvac` still supports Python2, for now, however:
* Dropping of Py2 support is planned: https://github.com/hvac/hvac/issues/582
* `hvac`'s own CI hasn't been testing against Python 2 since Feb 1: https://github.com/hvac/hvac/issues/582#issuecomment-771010264
* And that's led to two recent releases that accidentally broke Python 2 support, `0.10.12` and `0.10.13` (see https://github.com/hvac/hvac/issues/706 and https://github.com/hvac/hvac/issues/711) 
* And those issues have in turn renewed discussion of explicitly dropping support in the library **rather imminently**: https://github.com/hvac/hvac/issues/706#issuecomment-845337174

Even if `hvac` does not explicitly drop support, it's clear the lack of testing against it is a problem, and given the other difficulty of continuing to support it, I feel announcing the end of support in this collection's next major version is prudent.

---

This post will be updated with more information as it becomes available.

briantist:
## List of things to check/change when the change is made
just trying to keep a running list of stuff in place so I don't forget. I'm sure more will turn up.

- github workflows
- `tests/config.yml` -- update the required versions (this will only work in 2.12+ but still)
- `tests/integrations/requirements.txt`
- `tests/unit/requirements.txt`
[...]

PR #177, opened 2021-11-04
Drop support for Python 2 & 3.5

briantist:
##### SUMMARY

Remove explicit support for Python 2.7 and 3.5 (2.6 was already unsupported).

Resolves: #81 

Not looking to explicitly break anything in those python versions, but we won't be testing against it or enforcing support in PRs. We _may_ accept PRs that look to make something that stopped working, work again in an unsupported python version, if the change is not regressive or otherwise looks reasonable. It will be a case by case basis.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

N/A

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/changelogs/fragments/177-drop-py2-3.5.yml
+++ b/changelogs/fragments/177-drop-py2-3.5.yml
@@ -1,3 +1,3 @@
 ---
 removed_features:
-  - community.hashi_vault collection - drop support for Python 2 and Python 3.5 (https://github.com/ansible-collections/community.hashi_vault/issues/81).
+  - drop support for Python 2 and Python 3.5 (https://github.com/ansible-collections/community.hashi_vault/issues/81).
--- a/changelogs/fragments/177-drop-py2-3.5.yml
+++ b/changelogs/fragments/177-drop-py2-3.5.yml
@@ -1,0 +1,3 @@
+---
+removed_features:
+  - community.hashi_vault collection - drop support for Python 2 and Python 3.5 (https://github.com/ansible-collections/community.hashi_vault/issues/81).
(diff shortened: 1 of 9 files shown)


felixfontein (approved):

changelogs/fragments/177-drop-py2-3.5.yml
```diff
@@ -0,0 +1,3 @@
+---
+removed_features:
+  - community.hashi_vault collection - drop support for Python 2 and Python 3.5 (https://github.com/ansible-collections/community.hashi_vault/issues/81).
```
briantist, line 3:
```suggestion
  - drop support for Python 2 and Python 3.5 (https://github.com/ansible-collections/community.hashi_vault/issues/81).
```

felixfontein:
recheck

tests/unit/plugins/module_utils/test_hashi_vault_helper.py
```diff
@@ -29,7 +29,6 @@ def vault_token_via_env(vault_token):
         yield
 
 
-@pytest.mark.skipif(sys.version_info < (2, 7), reason="Python 2.7 or higher is required.")
```
felixfontein, line 32: Don't you somehow need the skip_python fixture here (and in other places)? Or how does that work?

briantist: @felixfontein yes indeed! It's in an autouse fixture in `tests/unit/conftest.py` (which existed before this, I just never went back to take out this one thing). That fixture is updated in this PR to change the version.

PR #196, opened 2021-12-13
Add tests for `ldap` auth method

### LINKED ISSUE #117: ldap auth method needs tests
##### SUMMARY

units can be added for some basic level of tests, though there's not much to this auth method from our end.

For integration we could look into setting up an LDAP server, but I may instead try use an MMock solution to test our end of it.

briantist:
##### SUMMARY

Fixes #117 

Adds unit and integration tests for the `ldap` auth method.

##### ISSUE TYPE

- Test Pull Request

##### COMPONENT NAME

tests/

##### ADDITIONAL INFORMATION

There have never been tests for `ldap` auth method; with the addition of MMock in #193 , we can test this in a similar way, mocking the result from Vault without having to set up an LDAP server.

### DIFF
--- a/tests/integration/targets/auth_ldap/aliases
+++ b/tests/integration/targets/auth_ldap/aliases
@@ -1,0 +1,2 @@
+vault/auth/ldap
+context/target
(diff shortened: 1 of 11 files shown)


briantist:
well, it doesn't appear I'm going to get any review for this, and no user-facing changes, so merging

PR #199, opened 2021-12-25
Add `vault_login` lookup and module, and `vault_login_token` filter plugin

briantist:
##### SUMMARY

Adds `vault_login` as a lookup and a module, which is used to authenticate to Vault and return the login response. 

The resulting token can then be re-used with other content with the `token` auth method. This provides a way to make a single login call, rather than performing a new login with every call to content in this collection. Some auth types can be expensive, or require a person to respond to external prompts like MFA.

A `vault_login_token` filter is also added to more easily extract the token value from the resulting login response structure.

Other changes:
- [units] the fixtures and their loader have been moved to the `conftest.py` in the root of the units, so they can be used in tests other than the auth method tests.
- A new Filter guide is added to the docsite to provide documentation and examples for the new filter (and any future filters).

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

vault_read

##### ADDITIONAL INFORMATION

N/A


briantist:
Note: the failed limited docs build is not a real failure: the full docs build shows that it all works correctly. The reason for the failure is that the new docsite filter guide page uses RST references to the new plugins in this PR, but the limited build does not use non-docstie content, and so those references don't exist in that build.

[This is discussed in more detail (with possible resolutions) here](https://github.com/ansible-collections/community.hashi_vault/issues/138#issuecomment-1001058332), however those resolutions can not be made in this PR.

plugins/lookup/vault_login.py
```diff
@@ -0,0 +1,124 @@
...
+      If the token does not have the C(lookup-self) capability, this will fail. If I(token_validate=False), only the token value itself
+      will be returned in the structure."
+  extends_documentation_fragment:
+    - community.hashi_vault.connection
+    - community.hashi_vault.connection.plugins
+    - community.hashi_vault.auth
+    - community.hashi_vault.auth.plugins
+  options:
+    _terms:
+      description: This is unused and should be an empty string.
```
felixfontein, line 40: Why not just pass nothing instead of an empty string?

plugins/lookup/vault_login.py
```diff
@@ -0,0 +1,124 @@
...
+      type: str
+      required: True
+"""
+
+EXAMPLES = """
+- name: Set a fact with a lookup result
+  set_fact:
+    login_data: "{{ lookup('community.hashi_vault.vault_login', '', url='https://vault', auth_method='userpass', username=user, password=pwd) }}"
```
felixfontein, line 48: Like this:
```suggestion
    login_data: "{{ lookup('community.hashi_vault.vault_login', url='https://vault', auth_method='userpass', username=user, password=pwd) }}"
```

plugins/lookup/vault_login.py
briantist: Thank you, confirmed this works and updated in https://github.com/ansible-collections/community.hashi_vault/pull/199/commits/c95599bf48ed30b3f264ee6c4973e27abb4c14f3

plugins/lookup/vault_login.py
briantist: Thank you, confirmed this works and updated in https://github.com/ansible-collections/community.hashi_vault/pull/199/commits/c95599bf48ed30b3f264ee6c4973e27abb4c14f3

felixfontein (approved):

PR #203, opened 2021-12-30
docs build- fix repo name on scheduled runs

briantist:
##### SUMMARY

Related: #202 

The default value for the collection name didn't work on scheduled runs because it's not available. Handling via the script plugin.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

docs build

##### ADDITIONAL INFORMATION

```paste below

```

PR #204, opened 2021-12-30
docs build - try to fix PR closed with no change

briantist:
##### SUMMARY

The new docs build was skipping the build step in a PR when it was closed. Because of this, it did not know whether the original PR had made docs changes or not, resulting in it always posting a comment that the "changes" had been merged.

This PR changes the behavior so that the build always runs, that way we can know whether there were changes. That makes it a bit slower, but that's hardly a concern. On a positive note, it simplifies the conditionals needed for running dependent jobs.

Also threw in another fix that should've been in #203 .

And added artifact name as an output from the build-html action'

---

Results from #204 look good for fixing change detection on closed/merged PRs and showing the appropriate message.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

docs build

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/actions/docs/ansible-docs-build-html/action.yml
+++ b/.github/actions/docs/ansible-docs-build-html/action.yml
@@ -34,6 +34,9 @@ outputs:
   artifact-url:
     description: The URL to the artifact.
     value: ${{ steps.outs.outputs.artifact-url }}
+  artifact-name:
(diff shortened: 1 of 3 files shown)


briantist:
did this PR wrong

PR #211, opened 2022-01-22
docs - lookup guide

briantist:
##### SUMMARY

Adds a lookup guide to the docsite. 
This discusses topics such as:
- when to use lookups vs. modules of the same name
- the effect of lazy templating on lookup invocation
- how/when/why lookups perform writes in Vault
- the `hashi_vault` lookup and its future in the collection

~This PR also includes the changes in https://github.com/ansible-community/antsibull/pull/391 since we commit the output of `antsibull sphinx-init` in the collection (so developers can build local docs more easily and so we can slightly customize the index page).~ 
^ these changes were moved to #215 .

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

docsite

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/docs/docsite/rst/lookup_guide.rst
+++ b/docs/docsite/rst/lookup_guide.rst
@@ -6,76 +6,15 @@ Lookup guide
 
 This guide is not a comprehensive listing of included lookup plugins and how to use them, rather it is intended to explain the role of the lookup plugins in ``community.hashi_vault`` and how they are they used, especially when compared to modules of the same name.
 
+For information about the ``hashi_vault`` lookup specifically, see :ref:`this page that covers it in detail <ansible_collections.community.hashi_vault.docsite.hashi_vault_lookup>`.
+
 .. contents::
   :local:
   :depth: 2
 
 
-About the ``hashi_vault`` lookup
-================================
-
-The ``hashi_vault`` lookup plugin is the oldest Vault-related content in Ansible. It was included in pre-collections Ansible (<2.10). As a result, it's the most used plugin for Vault, and the one most people are familiar with.
-
-``hashi_vault`` lookup considerations
--------------------------------------
-
-Due to the history of the ``hashi_vault`` lookup plugin, it does many jobs. It is versatile, but sometimes unintuitive.
-
-The ``hashi_vault`` lookup plugin performs three main tasks:
-
-* authentication, taking parameters for various login types, performing a login, and acquiring a token with which it can make additional calls to Vault.
-* a generic read operation, which allows it to read any kind of Vault path, without having to be written with that type of path in mind.
-* transforming responses that look like ``kv2`` responses into simpler responses that resemble those from ``kv1``.
-
-Reading secrets is the most common use case, with the ``kv`` (key/value) store built into Vault as by far the most common secret store. Most implementations use v2 of the ``kv`` store. To make reading v2 ``kv`` secrets easy, the lookup plugin assumes that you're probably trying to read a ``kv`` secret, and tries to infer if the response is from ``kv2``, because the responses from version 2 include metadata and have the secret value additionally wrapped in another structure. The lookup plugin seeks to make ``kv2`` responses look more like responses from version 1.
-
-Since the ``kv`` store has one or more key/value pairs in each secret, the lookup also supports a non-standard suffix in its path that can be used to access a value belonging to one specific key, via the ``:keyname`` syntax. While this is useful to provide a compact way to access a single secret value (admittedly a very common use case), it complicates the implementation and leads to bad habits.
-
-For example, it became common to see people use many lookup invocations with the same path, each with a different ``:keyname``, to access multiple values within a single secret, but this is quite wasteful, as it does a separate login and secret lookup, all to return the same value, and the key dereferencing is done client side. Further, dereferencing can be done directly in Jinja where it's more clear what's going on, using the ``.key`` or ``['key']`` syntax.
-
-One last idiosyncrasy of the plugin is its support for supplying all of its parameters in the term string. This looks compact, but it greatly complicates the processing of plugin options. At the time that this lookup was created, many other lookups allowed options to be supplied in the term string, but it has since been considered an anti-pattern, and has been deprecated/removed from core plugins.
-
-Another downside of this is that it prevents us from effectively re-using the authentication token in cases when multiple term strings are supplied, directly or via ``with_community.hashi_vault.hashi_vault``, and as a result this type of usage results in a new login for each term. In newer lookups, we can take advantage of a single login to perform multiple operations.
-
-All of these considerations make sense in context, but it somewhat muddles the purpose of the lookup:
-
-* If a response from a completely different endpoint ended up looking like a ``kv2`` response, it would return an unexpected result.
-* If you try to give the path of a ``kv2`` secret directly, it will not work unless you insert a ``/data/`` component into the path, in order to match the *API path* rather than the path people are usually familiar with.
-* If you want the metadata returned along with a ``kv2`` response, you cannot get it.
-* Other features of ``kv2`` like secret versioning cannot directly be used, unless you modify the URL, which is error prone and unintuitive.
-* Getting access to the token created by the internal login, in order to re-use it, is not possible.
-
-How we are addressing the considerations
-----------------------------------------
-
-The built-in authentication support will be kept, and in fact it has been moved to shared utilities within the collection, so that all plguins and modules can share the functionality, and work consistently. That makes it easier to test new and existing auth methods, easier to add new ones (which automaticallly become part of all existing content), and easier to add new content, because authentication does not need to be reimplemented.
-
-In addition, it is now possible to perform a login directly and return the token, for general re-use, via the ``community.hashi_vault.vault_login`` :ref:`module <ansible_collections.community.hashi_vault.vault_login_module>` and :ref:`lookup plugin <ansible_collections.community.hashi_vault.vault_login_lookup>`.
-
-Generic read (not ``kv`` specific) is still important functionality, so we have the ``community.hashi_vault.vault_read`` :ref:`module <ansible_collections.community.hashi_vault.vault_read_module>` and :ref:`lookup plugin <ansible_collections.community.hashi_vault.vault_read_lookup>` to provide that without trying to infer whether the response is from a specific backend.
-
-Since reading from ``kv`` store is by far the most common use case, we will also be introducing content for that specifically, which will accept familiar paths and provide parameters for ``kv``-specific functionality like versioning. That content is coming soon.
-
-The dictionary dereferencing via ``:keyname`` syntax *will not be supported* in other content. That will be achieved in Jinja via:
-
-* dot syntax ``.keyname``
-* lookup syntax ``['keyname']``
-* specialized filters in some circumstances, such as the ``vault_login_token`` :ref:`filter <ansible_collections.community.hashi_vault.docsite.filter_guide.vault_login_token>`.
-
-Parameters via term string *will not be supported* in other lookups. Its use is discouraged by core developers, and steps have already been taken in core to remove the functionality where it still exists, however it will remain in the ``hashi_vault`` plugin for backwards compatibility and because it is likely to still be in use in a lot of places.
-
-The future of the ``hashi_vault`` lookup
-----------------------------------------
-
-There are no plans currently to deprecate or remove the ``hashi_vault`` plugin. It is likely that it will stay indefinitely, for backwards compatibility and because so much functionality has been moved to shared code that very little maintenance is required to keep it. This decision may be revisited if circumstances change.
-
-That being said, we will encourage the use of newer content that has functionality with a tighter scope and is expected to receive updates and enchancements as appropriate.
-
-New features and functionality are unlikely to be added or accepted in the ``hashi_vault`` lookup, except for the ones that come for "free", like new auth methods (these require no code changes to the plugin itself).
-
 Lookups and writes
 ==================
-
 
 Most Ansible lookups perform read-only, non-destructive operations. They are run in templating, they generally *return*  values, and they **do not run differently in check mode** (that is they do the same thing they would in normal mode, even if that means changing something). However, some lookups do change state, sometimes by performing write operations. For example, the ``password`` :ref:`lookup <ansible_collections.ansible.builtin.password_lookup>` writes a generated password to a file, to act as a sort of cache, and the ``pipe`` :ref:`lookup <ansible_collections.ansible.builtin.pipe_lookup>` runs an arbitrary shell command so it could easily write or change state.
 
--- a/docs/docsite/rst/lookup_guide.rst
+++ b/docs/docsite/rst/lookup_guide.rst
@@ -119,7 +119,7 @@ Consider the following example:
       ansible.builtin.debug:
         msg: "Secret value A is '{{ value_a }}' while value B is '{{ value_b }}'."
 
-Since templating is recursive and evaluated lazily, this will unfortunately *not* result in a single login, resuing the token to perform a single secret read, which is then used is dictionary lookups.
+Since templating is recursive and evaluated lazily, this will unfortunately *not* result in a single login, reusing the token to perform a single secret read, which is then used is dictionary lookups.
 
 Instead, evaluation of ``value_a`` and ``value_b`` will *each* cause separate evaluation of ``secret``, so that lookup will be performed twice, and *each of those lookups* will cause a separate evaluation of ``token``, which will perform two separate logins, resulting in two tokens being created, and two reads of the exact same secret being performed.
 
--- a/docs/docsite/rst/lookup_guide.rst
+++ b/docs/docsite/rst/lookup_guide.rst
@@ -6,7 +6,7 @@ Lookup guide
 
 This guide is not a comprehensive listing of included lookup plugins and how to use them, rather it is intended to explain the role of the lookup plugins in ``community.hashi_vault`` and how they are they used, especially when compared to modules of the same name.
 
-For information about the ``hashi_vault`` lookup specifically, see :ref:`this page that covers it in detail <ansible_collections.community.hashi_vault.docsite.hashi_vault_lookup>`.
+For information about the ``hashi_vault`` lookup specifically, see :ref:`this page that covers it in detail <ansible_collections.community.hashi_vault.docsite.about_hashi_vault_lookup>`.
 
 .. contents::
   :local:
--- a/docs/docsite/rst/lookup_guide.rst
+++ b/docs/docsite/rst/lookup_guide.rst
@@ -23,9 +23,9 @@ Due to the history of the ``hashi_vault`` lookup plugin, it does many jobs. It i
 
 The ``hashi_vault`` lookup plugin performs three main tasks:
 
-- authentication, taking parameters for various login types, performing a login, and acquiring a token with which it can make additional calls to Vault.
-- a generic read operation, which allows it to read any kind of Vault path, without having to be written with that type of path in mind.
-- transforming responses that look like ``kv2`` responses into simpler responses that resemble those from ``kv1``.
+* authentication, taking parameters for various login types, performing a login, and acquiring a token with which it can make additional calls to Vault.
+* a generic read operation, which allows it to read any kind of Vault path, without having to be written with that type of path in mind.
+* transforming responses that look like ``kv2`` responses into simpler responses that resemble those from ``kv1``.
 
 Reading secrets is the most common use case, with the ``kv`` (key/value) store built into Vault as by far the most common secret store. Most implementations use v2 of the ``kv`` store. To make reading v2 ``kv`` secrets easy, the lookup plugin assumes that you're probably trying to read a ``kv`` secret, and tries to infer if the response is from ``kv2``, because the responses from version 2 include metadata and have the secret value additionally wrapped in another structure. The lookup plugin seeks to make ``kv2`` responses look more like responses from version 1.
 
--- a/docs/docsite/rst/lookup_guide.rst
+++ b/docs/docsite/rst/lookup_guide.rst
@@ -4,7 +4,7 @@
 Lookup guide
 ************
 
-This guide is not a comprehensive listing of included lookup plugins and how to use them, rather it is intended to explain the role of the lookup plugins in this collection and how they are they used, especially when compared to modules of the same name.
+This guide is not a comprehensive listing of included lookup plugins and how to use them, rather it is intended to explain the role of the lookup plugins in ``community.hashi_vault`` and how they are they used, especially when compared to modules of the same name.
 
 .. contents::
   :local:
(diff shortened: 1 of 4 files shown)


briantist:
recheck

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+    - vars:
+        token: "{{ lookup('community.hashi_vault.vault_login', auth_method='userpass', username='user', password='pass') | community.hashi_vault.vault_login_token }}"
+        secret: "{{ lookup('community.hashI_vault.vault_read', 'secrets/data/my-secret', token=token) }}"
+        value_a: "{{ secret.data.data.a }}"
+        value_b: "{{ secret.data.data.b }}"
+      ansible.builtin.debug:
+        msg: "Secret value A is '{{ value_a }}' while value B is '{{ value_b }}'."
+
+Since templating is recursive and evaluated lazily, this will unfortunately *not* result in a single login, resuing the token to perform a single secret read, which is then used is dictionary lookups.
```
felixfontein, line 122:
```suggestion
Since templating is recursive and evaluated lazily, this will unfortunately *not* result in a single login, reusing the token to perform a single secret read, which is then used is dictionary lookups.
```

felixfontein (commented):
I only skimmed some parts, but I found a typo while doing that:

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Due to the history of the ``hashi_vault`` lookup's development, and it being the only one around, it kind of does many jobs, making it versatile, but sometimes unintuitive.
+
+First, the lookup does authentication, taking parameters for various login types, performing a login, and acquiring a token with which it can make additional calls to Vault.
+
+Second, the lookup does a generic read. That makes it very useful because it can read any kind of Vault path, without having to be written with that type of path in mind.
+
+It's worth understanding that by far the most common use case is reading secrets, and by far the most common secret store is the ``kv`` (key/value) store that's built into Vault. Most implementations are using v2 of the ``kv`` store.
+
+So the third thing the lookup does is assume that you're probably trying to read a ``kv`` secret, and to then try to infer if the response is from ``kv2``, because the responses from version 2 include metadata and have the secret value additionally wrappped in another structure. In that way, it seeks to make the response from version 2 look more like the response from version 1.
```
acozine, line 30: I'd probably shorten this and turn it into a list:
```suggestion
The ``hashi_vault`` lookup plugin performs three main tasks:

 - authentication, taking parameters for various login types, performing a login, and acquiring a token with which it can make additional calls to Vault.
- a generic read operation, which allows it to read any kind of Vault path, without having to be written with that type of path in mind.
- reading secrets.

Reading secrets is the most common use case, with the ``kv`` (key/value) store built into Vault as by far the most common secret store. Most implementations use v2 of the ``kv`` store. To make reading v2 ``kv`` secrets easy, the lookup plugin assumes that you're probably trying to read a ``kv`` secret, and tries to infer if the response is from ``kv2``, because the responses from version 2 include metadata and have the secret value additionally wrapped in another structure. The lookup plugin seeks to make ``kv2`` responses look more like responses from version 1.
```

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Since the ``kv`` store has one or more key/value pairs in each secret, the lookup also supports a non-standard suffix in its path that can be used to access a value belonging to one specific key, via the ``:keyname`` syntax. While this is useful to provide a compact way to access a single secret value (admittedly a very common use case), it complicates the implementation and leads to bad habits.
+
+For example, it became common to see people use many lookup invocations with the same path, each with a different ``:keyname``, to access multiple values within a single secret, but this is quite wasteful, as it does a separate login and secret lookup, all to return the same value, and the key dereferencing is done client side. Further, dereferencing can be done directly in Jinja where it's more clear what's going on, using the ``.key`` or ``['key']`` syntax.
```
acozine, line 34: An example would be very helpful here, especially a "before and after" type of example, showing the same series of lookups, first done the "bad habit" way, then done the better way.

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+One last idiosyncrasy of the plugin is its support for supplying all of its parameters in the term string. This looks compact, but it greatly complicates the processing of plugin options. At the time that this lookup was created, many other lookups allowed options to be supplied in the term string, but it has since been considered an anti-pattern, and has been deprecated/removed from core plugins.
```
acozine, line 36: Here too a header and an example would be great.

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Another downside of this is that it prevents us from effectively re-using the authentication token in cases when multiple term strings are supplied, directly or via ``with_community.hashi_vault.hashi_vault``, and as a result this type of usage results in a new login for each term. In newer lookups, we can take advantage of a single login to perform multiple operations.
+
+All of these considerations make sense in context, but it somewhat muddles the purpose of the lookup:
```
acozine, line 40: I'm not entirely sure which of the previous things are included in "all of these considerations" here.

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+* If a response from a completely different endpoint ended up looking like a ``kv2`` response, it would return an unexpected result.
+* If you try to give the path of a ``kv2`` secret direcrly, it will not work unless you insert a ``/data/`` component into the path, in order to match the *API path* rather than the path people are usually familiar with.
+* If you want the metadata returned along with a ``kv2`` response, you cannot get it.
+* Other features of ``kv2`` like secret versioning cannot directly be used, unless you modify the URL, which is error prone and unintuitive.
+* Getting access to the token created by the internal login, in order to re-use it, is not possible.
+
+How we are addressing the considerations
```
acozine, line 48: Hmmmm. This is the moment where the doc changes from a description of what the plugin does, to a development roadmap of sorts. If that is the overall intention of the document, then I'd suggest a different main heading at the top and a slightly revised structure.

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+The future of the ``hashi_vault`` lookup
+----------------------------------------
+
+There are no plans currently to deprecate or remove the ``hashi_vault`` plugin. It is likely that it will stay indefinitely, for backwards compatibility and because so much functionality has been moved to shared code that very little maintenance is required to keep it. This decision may be revisited if circumstances change.
+
+That being said, we will encourage the use of newer content that has functionality with a tighter scope and is expected to receive updates and enchancements as appropriate.
+
+New features and functionality are unlikely to be added or accepted in the ``hashi_vault`` lookup, except for the ones that come for "free", like new auth methods (these require no code changes to the plugin itself).
+
+Lookups and writes
```
acozine, line 76: Now it feels like the doc is returning to more of a "what the lookup does and how to use it" doc. Maybe this is really two different documents, or one document with two major sections?

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+==================
+
+Lookups in Ansible kind of imply a read-only/non-destructive operation. Being run in templating, the expected result is that they *return* a value, and check mode has no effect on lookups. But there is no actual restriction on what a lookup can do, and some lookups do perform writes or otherwise change some state. For example, the ``password`` :ref:`lookup <ansible_collections.ansible.builtin.password_lookup>` writes a generated password to a file, to act as a sort of cache, and the ``pipe`` :ref:`lookup <ansible_collections.ansible.builtin.pipe_lookup>` runs an arbitrary shell command so it could easy write or change state.
```
acozine, line 79: Okay, I'm not sure I'm following this - but I think what you're saying is:
```suggestion

Most Ansible lookups perform read-only, non-destructive operations. They are run in templating, they generally *return*  values, and they do not generally run in check mode. However, some lookups do change state, sometimes by performing write operations. For example, the ``password`` :ref:`lookup <ansible_collections.ansible.builtin.password_lookup>` writes a generated password to a file, to act as a sort of cache, and the ``pipe`` :ref:`lookup <ansible_collections.ansible.builtin.pipe_lookup>` runs an arbitrary shell command so it could easy write or change state. The ``hashi_vault`` lookup also performs write operations.
```

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Writes in Vault
+---------------
+
+Operations that perform writes in Vault are not limited to the obvious ones such as writing a secret value, creating a policy, or enabling a new auth method.
```
acozine, line 84: Is this section more about Vault generally, or about ``hashi_vault`` lookup operations specifically?

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+When considering built-in support for auth methods, any auth method other than ``token`` or ``none`` makes every lookup, even ``vault_read``, into something that's changing state and performing a write within Vault. This actually applies to many modules too, even when using check mode.
+
+How to reason about when to use lookups
+---------------------------------------
+
+Because there is potential for writes in any lookup, it is very important to carefully consider when you are using a lookup vs. a module/other plugin. Check mode has no effect on lookups, so there is potential to perform many writes within your check mode run, but maybe sometimes you want that, for example if you're performing a ``vault_login`` via lookup to retrieve a token to use in your module calls, you may want that to still happen in check mode so that your module calls can properly check the things they need to.
```
acozine, line 101: Ahhhh, I see what you mean by "check mode has no effect on lookups" - you mean, "even running in check mode, your lookup plugin is going to do whatever it would do on a real run, so be careful . . . " That is a very important point - I'd make it more explicit here and above.

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Some modules that are read focused, like the ``vault_read`` module, when used with auth other than ``token`` or ``none``, will still perform an internal login even in check mode, so this is still another consideration.
+
+Lookups and lazy templating
+---------------------------
+
+The capacity for lookups to perform writes or change state is exacerbated by Ansible's "lazy" templating, if not used carefully.
+
+Consider the following example:
```
acozine, line 110: Hooray for examples!

acozine (commented):
There's a lot of really great material in this PR, including a development roadmap, helpful warnings, and general how-to documentation about the `hashi_vault` lookup itself. I've tried to suggest ways to organize it to bring out all the gems included here.

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Second, the lookup does a generic read. That makes it very useful because it can read any kind of Vault path, without having to be written with that type of path in mind.
+
+It's worth understanding that by far the most common use case is reading secrets, and by far the most common secret store is the ``kv`` (key/value) store that's built into Vault. Most implementations are using v2 of the ``kv`` store.
+
+So the third thing the lookup does is assume that you're probably trying to read a ``kv`` secret, and to then try to infer if the response is from ``kv2``, because the responses from version 2 include metadata and have the secret value additionally wrappped in another structure. In that way, it seeks to make the response from version 2 look more like the response from version 1.
+
+Since the ``kv`` store has one or more key/value pairs in each secret, the lookup also supports a non-standard suffix in its path that can be used to access a value belonging to one specific key, via the ``:keyname`` syntax. While this is useful to provide a compact way to access a single secret value (admittedly a very common use case), it complicates the implementation and leads to bad habits.
```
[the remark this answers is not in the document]
briantist: Maybe it would be better just to shorten it, I don't want to call too much attention to it since we're not looking to promote use of it, it was meant as historical context basically. Will think about it a bit.

docs/docsite/rst/lookup_guide.rst
briantist: I actually do intend to move a bunch of this to a new page in the future, but I wanted to put some of it down somewhere for now.

What I'm waiting for for the other page, is the expected `vault_kv_` series of plugins and modules, which together with `vault_read` will provide full replacement coverage for `hashi_vault`, by separating the functionality into "reading kv secrets" and "reading anything else".

At that point, I'd have a guide specifically for migrating away from using `hashi_vault` and what to use in its place, that would be full of examples and such.

---

So this paragraph, could maybe be shortened or removed entirely (I don't necessarily need to explain what the bad habits are here).

docs/docsite/rst/lookup_guide.rst
briantist: I am leaning toward just removing or severely cutting the text down instead. I would usually agree with adding headings, but I think the issue may be that this entire section of historical context is just too long.

docs/docsite/rst/lookup_guide.rst
briantist: well.. all of the previous things, from the beginning of the section (which in retrospect, is too many things, or too much explanation of a few things) 😅

docs/docsite/rst/lookup_guide.rst
briantist: It isn't intended to be a roadmap, only to address the question of this plugin that we will kind of be encouraging folks to consider not using in favor of alternatives.

docs/docsite/rst/lookup_guide.rst
briantist: Yeah, you've hit the nail on the head, the stuff about `hashi_vault` is really intended to go into a dedicated page about alternatives (newer content in the collection), but I was waiting to fully write that page until the release of `vault_kv_` content. As I was writing this page, I thought it might be a good opportunity to start talking about some of that.. maybe I should remove the whole section completely and either wait, or post it as a separate page, even while acknowledging that there's no direct `kv` alternative right now.

docs/docsite/rst/lookup_guide.rst
briantist: Yeah your rewrite is correct, with a few small details I can update:
* lookups _do_ run in check mode (as in, they perform whatever they were going to do, blissfully unaware that your intention was to run ansible in a way that doesn't make changes)
* this section isn't about the `hashi_vault` plugin, even though it along with the other lookups in the collection do perform writes of a sort, implicitly. This is just a brief overview/reminder(?) of the way lookups work and that they imply read _by convention_ but are not limited to that.

Ultimately this, and the next section (about writes in Vault generally), are the pre-requisites to understanding how to properly use lookups in a way that you are not surprised by the behavior, and can understand when things that don't look like writes can have an effect at scale (if you run tens of thousands of lookups that all seem like reads to you, you might be surprised when Vault runs into storage/replication issues due to a sudden flood of token creation for example).

docs/docsite/rst/lookup_guide.rst
briantist: This section is about Vault generally, and then relating those things to the lookups in this collection (not just the `hashi_vault` lookup).

docs/docsite/rst/lookup_guide.rst
briantist: Yes exactly. I may move some of this stuff about writes and implicit writes into its own document at some point as well, because while it can be trickier with lookups, some of it applies to modules too. 

For example: `vault_login` (the module) does not perform a login in check mode. `vault_read` (the module), takes all the same authentication parameters as `vault_login`, and will perform the login even in check mode, so that it can execute the read in check mode as well. The `vault_token_create` module (#213) similarly will perform the login with its authentication parameters, but will not create a token in check mode.

But for now I think I will like to keep some of this here, there's always room to reorganize and move things to new pages later.

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Another downside of this is that it prevents us from effectively re-using the authentication token in cases when multiple term strings are supplied, directly or via ``with_community.hashi_vault.hashi_vault``, and as a result this type of usage results in a new login for each term. In newer lookups, we can take advantage of a single login to perform multiple operations.
+
+All of these considerations make sense in context, but it somewhat muddles the purpose of the lookup:
+
+* If a response from a completely different endpoint ended up looking like a ``kv2`` response, it would return an unexpected result.
+* If you try to give the path of a ``kv2`` secret direcrly, it will not work unless you insert a ``/data/`` component into the path, in order to match the *API path* rather than the path people are usually familiar with.
```
briantist, line 43:
```suggestion
* If you try to give the path of a ``kv2`` secret directly, it will not work unless you insert a ``/data/`` component into the path, in order to match the *API path* rather than the path people are usually familiar with.
```

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
+.. _ansible_collections.community.hashi_vault.docsite.lookup_guide:
+
+************
+Lookup guide
+************
+
+This guide is not a comprehensive listing of included lookup plugins and how to use them, rather it is intended to explain the role of the lookup plugins in this collection and how they are they used, especially when compared to modules of the same name.
```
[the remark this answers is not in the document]
briantist:
```suggestion
This guide is not a comprehensive listing of included lookup plugins and how to use them, rather it is intended to explain the role of the lookup plugins in ``community.hashi_vault`` and how they are they used, especially when compared to modules of the same name.
```

docs/docsite/rst/lookup_guide.rst
briantist: this is much better thanks, I will have to reword slightly but I'll commit and edit after

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Due to the history of the ``hashi_vault`` lookup plugin, it does many jobs. It is versatile, but sometimes unintuitive.
+
+The ``hashi_vault`` lookup plugin performs three main tasks:
+
+ - authentication, taking parameters for various login types, performing a login, and acquiring a token with which it can make additional calls to Vault.
+- a generic read operation, which allows it to read any kind of Vault path, without having to be written with that type of path in mind.
+- reading secrets.
+
+Reading secrets is the most common use case, with the ``kv`` (key/value) store built into Vault as by far the most common secret store. Most implementations use v2 of the ``kv`` store. To make reading v2 ``kv`` secrets easy, the lookup plugin assumes that you're probably trying to read a ``kv`` secret, and tries to infer if the response is from ``kv2``, because the responses from version 2 include metadata and have the secret value additionally wrapped in another structure. The lookup plugin seeks to make ``kv2`` responses look more like responses from version 1.
```
briantist, line 30:
```suggestion
 - authentication, taking parameters for various login types, performing a login, and acquiring a token with which it can make additional calls to Vault.
- a generic read operation, which allows it to read any kind of Vault path, without having to be written with that type of path in mind.
- modifying the response from Vault to be more useful it came from a ``kv2`` backend.

Expanding on that last point: reading secrets is the most common use case, with the ``kv`` (key/value) store built into Vault as by far the most common secret store. Most implementations use v2 of the ``kv`` store. To make reading v2 ``kv`` secrets easy, the lookup plugin assumes that you're probably trying to read a ``kv`` secret, and tries to infer if the response is from ``kv2``, because the responses from version 2 include metadata and have the secret value additionally wrapped in another structure. In short: the ``hashi_vault`` lookup seeks to make ``kv2`` responses look more like the (simpler) responses from version 1.
```

briantist: @acozine I committed your previous suggestion and then made a few changes onto it, lmk what you think.

acozine: Nice!

docs/docsite/rst/lookup_guide.rst
acozine: Maybe we can square the circle here by putting a sentence into the introduction . . . something like "The `hashi_vault` lookup plugin can perform many module-like actions. In future releases of the `community.hashi_vault` collection, new module plugins will perform these actions, making Ansible Vault interactions more predictable and more parallel with other Ansible behaviors."

docs/docsite/rst/lookup_guide.rst
acozine: Ah, gotcha.

docs/docsite/rst/lookup_guide.rst
briantist, line 30:
```suggestion
The ``hashi_vault`` lookup plugin performs three main tasks:

 - authentication, taking parameters for various login types, performing a login, and acquiring a token with which it can make additional calls to Vault.
- a generic read operation, which allows it to read any kind of Vault path, without having to be written with that type of path in mind.
- transforming responses that look like ``kv2`` responses into simpler responses that resemble those from ``kv1``.

Reading secrets is the most common use case, with the ``kv`` (key/value) store built into Vault as by far the most common secret store. Most implementations use v2 of the ``kv`` store. To make reading v2 ``kv`` secrets easy, the lookup plugin assumes that you're probably trying to read a ``kv`` secret, and tries to infer if the response is from ``kv2``, because the responses from version 2 include metadata and have the secret value additionally wrapped in another structure. The lookup plugin seeks to make ``kv2`` responses look more like responses from version 1.
```

docs/docsite/rst/lookup_guide.rst
```diff
@@ -0,0 +1,160 @@
...
+
+That being said, we will encourage the use of newer content that has functionality with a tighter scope and is expected to receive updates and enchancements as appropriate.
+
+New features and functionality are unlikely to be added or accepted in the ``hashi_vault`` lookup, except for the ones that come for "free", like new auth methods (these require no code changes to the plugin itself).
+
+Lookups and writes
+==================
+
+
+Most Ansible lookups perform read-only, non-destructive operations. They are run in templating, they generally *return*  values, and they do not generally run in check mode. However, some lookups do change state, sometimes by performing write operations. For example, the ``password`` :ref:`lookup <ansible_collections.ansible.builtin.password_lookup>` writes a generated password to a file, to act as a sort of cache, and the ``pipe`` :ref:`lookup <ansible_collections.ansible.builtin.pipe_lookup>` runs an arbitrary shell command so it could easy write or change state. The ``hashi_vault`` lookup also performs write operations.
```
briantist, line 80:
```suggestion
Most Ansible lookups perform read-only, non-destructive operations. They are run in templating, they generally *return*  values, and they **do not run differently in check mode** (that is they do the same thing they would in normal mode, even if that means changing something). However, some lookups do change state, sometimes by performing write operations. For example, the ``password`` :ref:`lookup <ansible_collections.ansible.builtin.password_lookup>` writes a generated password to a file, to act as a sort of cache, and the ``pipe`` :ref:`lookup <ansible_collections.ansible.builtin.pipe_lookup>` runs an arbitrary shell command so it could easily write or change state. 
```

briantist:
@acozine thanks so much for your help on this, your suggestions so far have really improved it already! I merged today so that I could get this content into a release, but I'm happy to accept PRs to it, or if you have comments on the closed PR _I will read them_ and make adjustments. Thanks again!

Issue #210, opened 2022-01-19
hashi_vault_lookup create token / orphan token


emper0r:
##### ISSUE TYPE
- How could be possible create a token and/or orphan token with this module???
according to the doc https://www.vaultproject.io/docs/concepts/tokens#token-hierarchies-and-orphan-tokens
create a token do a POST using write function into endpoint and lookup just do GET.
in this case isn't "write" a data .... we are asking to Vault to give us a token made by itself so would be "like a GET method" but... API for this need a POST.

##### COMPONENT NAME

hashi_vault_lookup

##### ANSIBLE VERSION

```paste below
4.3.0
```

##### COLLECTION VERSION

```paste below
1.1.3 (using today)
2.2.0 (test it)
```

##### CONFIGURATION

```paste below
not exist.... now we are doing workaround using API with URI module

- name: main | Create token
  ansible.builtin.uri:
    headers:
      X-Vault-Token: "{{ lookup('env', 'VAULT_TOKEN') }}"
    url: "{{ env_vault_addr }}/v1/auth/token/create-orphan"
    method: POST
    body_format: json
    status_code: 200
    return_content: true

```

##### STEPS TO REPRODUCE

```yaml

```

##### EXPECTED RESULTS

Concept or an idea about it
```
new_orphan_token: "{{ lookup('community.hashi_vault.hashi_vault', token="new'', token_type=" ( periodic | orphan ), path="auth/token/create-orphan", ..... ) }}"
```
##### ACTUAL RESULTS

```paste below

```

briantist:
Hi @emper0r , the latest released version of `community.hashi_vault`, ver `2.2.0`, contains a lookup plugin and a module called `vault_login`:
- lookup: https://docs.ansible.com/ansible/devel/collections/community/hashi_vault/vault_login_lookup.html
- module: https://docs.ansible.com/ansible/devel/collections/community/hashi_vault/vault_login_lookup.html

However these don't seem to do what you'd like, which is to use an existing token (authorization) to then create a new (possibly orphaned) token.

I do plan to add that capability in the collection, along with many other capabilities, but I do not have a timeline just yet.

I do like to respond to user requests when possible though, so because you requested it, I may implement something sooner than expected; I will be sure to update this issue with any news.

emper0r:
[...]

PR #213, opened 2022-01-30
`vault_token_create` - add module and lookup for creating tokens

briantist:
##### SUMMARY

Resolves: #210

Adds `vault_token_create` as a module and lookup plugin. Supports creating orphan tokens as well, but the support for that in `hvac` is in a weird state. The deprecated method for creating tokens supports making calls to the `/token/create-orphan` endpoint, but its replacement can only make calls to `/token/create`, and that endpoint requires root ~or sudo~ privileges to create orphans, while the latter ~does not~ requires `sudo` but not `root`. See also:
- https://github.com/hvac/hvac/issues/758

We're handling this by having an `orphan` boolean option in addition to the `no_parent` option, just as the deprecated method did, and when it's `true`, we will use the deprecated method for now. If deprecated method is unavailable, a last ditch effort to call the new method will be made, but that will only succeed if you're making the call with a root token.

We will see how the linked issue develops in `hvac`.

Other trivial changes included here:
- some copyright years were meant to be 2021 but typo'd to 2012 (fixed)
- the `_terms` option in the `vault_login` lookup was marked as required even though that is not the case (in fact anything passed in will be ignored); that has been fixed.
- added stub for missing `vault_read` lookup unit tests, opened #214 
- added `integration.cfg` (follow up: #218)
- fixed `vault_ci_token_create`'s `policies` option, it was supposed to be a list but was a string.
- updated the filter guide to mention these new plugins, since the `vault_login_token` filter can be used with these too.
- tried to ignore `conftest.py` files in `codecov.yml`, but it's not reflected in the PR report, I guess we'll see if it worked upon merging.

Non-trivial things:
- a nice `conftest.py` for unit testing modules, based on one floating around in core/collections
- the first unit tests for a module, and they're nice!

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

vault_token_create


briantist:
@emper0r this is ready for review, if you  have some time to take a look and try it out.

plugins/doc_fragments/token_create.py
```diff
@@ -0,0 +1,98 @@
...
+class ModuleDocFragment(object):
+
+    DOCUMENTATION = r'''
+options:
+  orphan:
+    description:
+      - When C(true), uses the C(/create-orphan) API endpoint, which requires C(sudo) (but not C(root)) to create an orphan.
+      - Implies I(no_parent=true).
+      - B(NOTE:) as of this writing, the underlying endpoint in the C(hvac) library to support this is deprecated and scheduled for removal in C(v1.0.0).
+      - If I(orphan=true) and we cannot access the intended enpoint, the call will be attempted with the C(/create) endpoint, which requires root.
```
Andersson007, line 19:
```suggestion
      - If I(orphan=true) and we cannot access the intended endpoint, the call will be attempted with the C(/create) endpoint, which requires root.
```

plugins/doc_fragments/token_create.py
```diff
@@ -0,0 +1,98 @@
...
+    description: The token type. The default is determined by the role configuration specified by I(role_name).
+    type: str
+    choices:
+      - batch
+      - service
+  explicit_max_ttl:
+    description:
+      - If set, the token will have an explicit max TTL set upon it.
+      - "This maximum token TTL cannot be changed later,
+        and unlike with normal tokens, updates to the system/mount max TTL value will have no effect at renewal time."
```
Andersson007, line 72:
```suggestion
      - This maximum token TTL cannot be changed later,
        and unlike with normal tokens, updates to the system/mount max TTL value will have no effect at renewal time.
```

plugins/lookup/vault_token_create.py
```diff
@@ -0,0 +1,176 @@
...
+    - module: community.hashi_vault.vault_login
+    - ref: community.hashi_vault.vault_login_token filter <ansible_collections.community.hashi_vault.docsite.filter_guide.vault_login_token>
+      description: The official documentation for the C(community.hashi_vault.vault_login_token) filter plugin.
+  notes:
+    - Token creation is a write operation (creating a token persisted to storage), so this module always reports C(changed=True).
+    - "For the purposes of Ansible playbooks however,
+      it may be more useful to set C(changed_when=false) if you're doing idempotency checks against the target system."
+    - "In check mode, this module will not create a token, and will instead return a basic structure with an empty token.
+      However this may not be useful if the token is required for follow on tasks.
+      It may be better to use this module with C(check_mode=no) in order to have a valid token that can be used."
```
Andersson007, line 31:
```suggestion
    - For the purposes of Ansible playbooks however,
      it may be more useful to set I(changed_when=false) if you are doing idempotency checks against the target system.
    - In check mode, this module will not create a token, and will instead return a basic structure with an empty token.
      However, this may not be useful if the token is required for follow on tasks.
      It may be better to use this module with I(check_mode=no) in order to have a valid token that can be used.
```
about using `you're` -> `you are` - it's from the Ansible doc style guide.

plugins/lookup/vault_token_create.py
```diff
@@ -0,0 +1,176 @@
...
+  short_description: Create a HashiCorp Vault token
+  requirements:
+    - C(hvac) (L(Python library,https://hvac.readthedocs.io/en/stable/overview.html))
+    - For detailed requirements, see R(the collection requirements page,ansible_collections.community.hashi_vault.docsite.user_guide.requirements).
+  description:
+    - Creates a token in HashiCorp Vault, returning the response, including the token.
+  seealso:
+    - module: community.hashi_vault.vault_token_create
+    - ref: community.hashi_vault.vault_login lookup <ansible_collections.community.hashi_vault.vault_login_lookup>
+      description: The official documentation for the C(community.hashi_vault.vault_login) lookup plugin.
```
Andersson007, line 21:
```suggestion
      description: The official documentation for the M(community.hashi_vault.vault_login) lookup plugin.
```

plugins/lookup/vault_token_create.py
```diff
@@ -0,0 +1,176 @@
...
+    - module: community.hashi_vault.vault_login
+    - ref: community.hashi_vault.vault_login_token filter <ansible_collections.community.hashi_vault.docsite.filter_guide.vault_login_token>
+      description: The official documentation for the C(community.hashi_vault.vault_login_token) filter plugin.
```
Andersson007, line 24:
```suggestion
      description: The official documentation for the M(community.hashi_vault.vault_login_token) filter plugin.
```

plugins/lookup/vault_token_create.py
```diff
@@ -0,0 +1,176 @@
...
+  options:
+    _terms:
+      description: This is unused and any terms supplied will be ignored.
+      type: str
+      required: false
+"""
+
+EXAMPLES = """
+- name: Login via userpass and create a child token
+  set_fact:
```
Andersson007, line 47:
```suggestion
  ansible.builtin.set_fact:
```

plugins/modules/vault_token_create.py
```diff
@@ -0,0 +1,213 @@
...
+    path: auth/approle/role/role-name/role-id
+  register: approle_id
+
+- name: Retrieve an approle role ID using the child token (token via direct dict access)
+  community.hashi_vault.vault_read:
+    url: https://vault:8201
+    auth_method: token
+    token: '{{ token_data.login.auth.client_token }}'
+    path: auth/approle/role/role-name/role-id
+  register: approle_id
```
Andersson007, line 65:
```suggestion
  register: approle_id
```
we could add a debug task for this variable as we do a bit below for token_data.login

plugins/modules/vault_token_create.py
```diff
@@ -0,0 +1,213 @@
...
+    ttl: 60s
+  register: token_data
+
+- name: Display the full response
+  ansible.builtin.debug:
+    var: token_data.login
+"""
+
+RETURN = """
+login:
```
Andersson007, line 81:
```suggestion
login:
```
should we also add `sample:` ?

plugins/modules/vault_token_create.py
```diff
@@ -0,0 +1,213 @@
...
+    - Brian Scholer (@briantist)
+  short_description: Create a HashiCorp Vault token
+  requirements:
+    - C(hvac) (L(Python library,https://hvac.readthedocs.io/en/stable/overview.html))
+    - For detailed requirements, see R(the collection requirements page,ansible_collections.community.hashi_vault.docsite.user_guide.requirements).
+  description:
+    - Creates a token in HashiCorp Vault, returning the response, including the token.
+  seealso:
+    - ref: community.hashi_vault.vault_token_create lookup <ansible_collections.community.hashi_vault.vault_token_create_lookup>
+      description: The official documentation for the C(community.hashi_vault.vault_token_create) lookup plugin.
```
Andersson007, line 22:
```suggestion
      description: The official documentation for the M(community.hashi_vault.vault_token_create) lookup plugin.
```

plugins/modules/vault_token_create.py
```diff
@@ -0,0 +1,213 @@
...
+    - module: community.hashi_vault.vault_login
+    - ref: community.hashi_vault.vault_login lookup <ansible_collections.community.hashi_vault.vault_login_lookup>
+      description: The official documentation for the C(community.hashi_vault.vault_login) lookup plugin.
+    - ref: community.hashi_vault.vault_login_token filter <ansible_collections.community.hashi_vault.docsite.filter_guide.vault_login_token>
+      description: The official documentation for the C(community.hashi_vault.vault_login_token) filter plugin.
```
Andersson007, line 27:
```suggestion
      description: The official documentation for the M(community.hashi_vault.vault_login_token) filter plugin.
```

plugins/modules/vault_token_create.py
```diff
@@ -0,0 +1,213 @@
...
+    - community.hashi_vault.connection
+    - community.hashi_vault.auth
+    - community.hashi_vault.token_create
+  notes:
+    - Token creation is a write operation (creating a token persisted to storage), so this module always reports C(changed=True).
+    - "For the purposes of Ansible playbooks however,
+      it may be more useful to set C(changed_when=false) if you're doing idempotency checks against the target system."
+    - "In check mode, this module will not create a token, and will instead return a basic structure with an empty token.
+      However this may not be useful if the token is required for follow on tasks.
+      It may be better to use this module with C(check_mode=no) in order to have a valid token that can be used."
```
Andersson007, line 38:
```suggestion
    - For the purposes of Ansible playbooks however,
      it may be more useful to set I(changed_when=false) if you are doing idempotency checks against the target system.
    - In check mode, this module will not create a token, and will instead return a basic structure with an empty token.
      However, this may not be useful if the token is required for follow on tasks.
      It may be better to use this module with I(check_mode=no) in order to have a valid token that can be used.
```

plugins/lookup/vault_token_create.py
```diff
@@ -0,0 +1,176 @@
...
+  register: approle_id
+
+# implicitly uses url & token auth with a token from the environment
+- name: Create an orphaned token with a short TTL and display the full response
+  ansible.builtin.debug:
+    var: lookup('community.hashi_vault.vault_token_create', orphan=True, ttl='60s')
+"""
+
+RETURN = """
+_raw:
```
Andersson007, line 73:
```suggestion
_raw:
```
how about putting `sample:` here?

Andersson007 (commented):
@briantist thanks for the enhancement!
Several mostly cosmetic things from me.

plugins/modules/vault_token_create.py
briantist: perhaps, but the debug task is more relevant to this module; retrieving the `approle_id` is a sample of one way to use the token, but the returned ID probably isn't something we need to look at as part of the examples here.. I dunno maybe will add it.

plugins/modules/vault_token_create.py
briantist: re: the samples, I could either add a minimal sample that covers only the things called out in the return, or an example of a full(ish) structure that's likely to be returned, which is much bigger (see the `token_create_response.json` fixture in this PR). What do you think?

plugins/lookup/vault_token_create.py
Andersson007: yeah, I've re-read the doc, you're right, sorry:)

plugins/lookup/vault_token_create.py
Andersson007: wrong suggestion

plugins/modules/vault_token_create.py
Andersson007: ah, ok, thanks for clarifying

plugins/modules/vault_token_create.py
Andersson007: do as you consider right, it's just a reminder:) maybe the minimal variant would be good

plugins/modules/vault_token_create.py
```diff
@@ -0,0 +1,213 @@
...
+  ansible.builtin.debug:
+    var: token_data.login
+"""
+
+RETURN = """
+login:
+  description: The result of the token creation operation.
+  returned: success
+  type: dict
+  contains:
```
briantist, line 85:
```suggestion
  sample:
    auth:
      client_token: s.rlwajI2bblHAWU7uPqZhLru3
    data: null
  contains:
```

plugins/lookup/vault_token_create.py
```diff
@@ -0,0 +1,176 @@
...
+  description: The result of the token creation operation.
+  returned: success
+  type: dict
+  contains:
```
briantist, line 77:
```suggestion
  sample:
    auth:
      client_token: s.rlwajI2bblHAWU7uPqZhLru3
    data: null
  contains:
```

Andersson007 (approved):
I'm unfamiliar with the related technical area but, at least in terms of doc and tests (that's great when things are covered by both unit and integration tests), LGTM

@briantist thank you for new modules!

briantist:
@Andersson007 thanks for reviewing!

PR #216, opened 2022-02-08
Add module `vault_pki_generate_certificate`

Ripolin:
##### SUMMARY

Add a new module to generate a certificate with Vault PKI engine (https://www.vaultproject.io/api/secret/pki#generate-certificate)

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

vault_generate_certificate

##### ADDITIONAL INFORMATION
Add a ```plugins/modules/pki``` package to dissociate pki and generic modules if more pki modules are developped (https://hvac.readthedocs.io/en/stable/usage/secrets_engines/pki.html#generate-certificate).


briantist:
Hi @Ripolin and welcome! It might take me some time to give this a proper review, but some early feedback:
- let's skip the subdirectory in `plugins/modules` and put the module right alongside the rest for now
- would prefer that the name is `vault_pki_generate_certificate`
- at a quick glance the module look good but it will need extensive test coverage to be accepted, please take a look at the other integration tests to get a sense of how to structure these tests

More info on integration tests:
- this is the first module we have that works (directly) against another secrets backend, so the backend will have to be set up in the integration test role
- for the time being, I think it's ok to do backend setup and teardown (removal) of the backend within the role; later on if there are more pki modules, we may separate that out into its own `setup_` target, but that can wait
- ensure that the teardown happens even when an error occurs in the role

Unit tests are nice but not yet required for modules and I don't have any examples (I am looking into adding unit tests for the module version of `vault_token_create` in #213 so there may be examples there soon).

briantist:
> @briantist Wip

Thank you! There is no need to force push your branch or squash commits, we squash on merge, so please free to pus up as many commits as needed. It can be helpful to see previous commits to understand how things changed over time.

Ripolin:
> > @briantist Wip
> 
> Thank you! There is no need to force push your branch or squash commits, we squash on merge, so please free to pus up as many commits as needed. It can be helpful to see previous commits to understand how things changed over time.

I know but I like the idea to just have one commit to merge, to me it's sound prettier :sweat_smile:

briantist:
> > > @briantist Wip
> > 
> > 
> > Thank you! There is no need to force push your branch or squash commits, we squash on merge, so please free to pus up as many commits as needed. It can be helpful to see previous commits to understand how things changed over time.
> 
> I know but I like the idea to just have one commit to merge, to me it's sound prettier 😅

Understood, but it is more helpful for me as a reviewer to have multiple commits to view for context, if needed, and regardless of the number of commits, on PR merge, there will be a single merge commit anyway.

You can continue to squash and force push if you like, but if the single merge is the only issue, you don't have to worry about that 😄

plugins/modules/vault_pki_generate_certificate.py
```diff
@@ -0,0 +1,266 @@
...
+# -*- coding: utf-8 -*-
+# (c) 2022, Florent David <florent.david@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+  module: vault_pki_generate_certificate
+  version_added: 2.11.6
```
briantist, line 11:
```suggestion
  version_added: 2.3.0
```
this needs to be collection version, and next expected for us is `2.3.0`

plugins/modules/vault_pki_generate_certificate.py
```diff
@@ -0,0 +1,266 @@
...
+      description: API documentation for the Hashicorp Vault PKI secrets engine.
+      link: https://www.vaultproject.io/api/secret/pki#generate-certificate
+    - name: HVAC library reference
+      description: Reference of the HVAC library about PKI engine.
+      link: https://hvac.readthedocs.io/en/stable/usage/secrets_engines/pki.html#generate-certificate
+  extends_documentation_fragment:
+    - community.hashi_vault.connection
+    - community.hashi_vault.auth
+  options:
+    - alt_names:
```
briantist, line 31: The option names should be keys in the dictionary of `options`, not list items, this is why the tests are failing currently, so:
```yaml
  options:
    alt_names:
      description: ...
    next_option:
      description: etc...
```

plugins/modules/vault_pki_generate_certificate.py
Ripolin: Fixed in commit 3a81a4ba8da0bef037f518023c3e4553e7c66cda

plugins/modules/vault_pki_generate_certificate.py
Ripolin: Fixed in commit 3a81a4ba8da0bef037f518023c3e4553e7c66cda

Ripolin:
> 



> By the way, I strongly recommend getting set up to run tests locally, you'll be able to iterate much faster not having to push and wait for CI, especially since as a new contributor, I will have to approve CI runs for each push.
> 
> We have a contributor guide that covers how to run the tests and if anything is unclear or you have any questions, please let me know! https://docs.ansible.com/ansible/devel/collections/community/hashi_vault/docsite/contributor_guide.html#running-tests-locally

This workaround works on Linux, but on Windows it seems harder to run. So I update integration tests and I'll try to run them soon before pushing it

Ripolin:
@briantist Integration tests works locally, I push them. However documentation still have problems : https://community-hashi-vault-pr216.surge.sh/collections/community/hashi_vault/vault_pki_generate_certificate_module.html. Any idea ?

briantist:
> > By the way, I strongly recommend getting set up to run tests locally, you'll be able to iterate much faster not having to push and wait for CI, especially since as a new contributor, I will have to approve CI runs for each push.
> > We have a contributor guide that covers how to run the tests and if anything is unclear or you have any questions, please let me know! https://docs.ansible.com/ansible/devel/collections/community/hashi_vault/docsite/contributor_guide.html#running-tests-locally
> 
> This workaround works on Linux, but on Windows it seems harder to run. So I update integration tests and I'll try to run them soon before pushing them

I run Windows as well, I use WSL (Ubuntu, but you can probably use other distros) for all my Ansible stuff, including the test suite for this collection. The easiest way is with docker, but it is also possible to run tests without docker, it will just be more involved. The sanity tests are easier to run since they don't even require Vault or the other containers we use in integration.

Do you happen to have WSL set up?

> @briantist Integration tests works locally so I push them. However documentation still have problems : https://community-hashi-vault-pr216.surge.sh/collections/community/hashi_vault/vault_pki_generate_certificate_module.html. Any idea ?

I approved the other CI and sanity has found several errors in the docs:
https://github.com/ansible-collections/community.hashi_vault/runs/5135271236?check_suite_focus=true

If you can get sanity running locally you should be able to fix those very quickly. I think once sanity passes the docs will probably build quickly, or at least we will get a better error if there is still a problem.

plugins/modules/vault_pki_generate_certificate.py
```diff
@@ -0,0 +1,269 @@
...
+          sample: 39:dd:2e:90:b7:23:1f:8d:d3:7d:31:c5:1b:da:84:d0:5b:65:31:58
+    warning:
+      description: Warning throw during generation
+      returned: success
+      type: string
+"""
+
+import traceback
+
+from ansible.errors import AnsibleError
```
briantist, line 192: Modules cannot import from Ansible core libraries except `module_utils`, so we can't use `AnsibleError` (this exception class is used in plugins).
```suggestion
```

plugins/modules/vault_pki_generate_certificate.py
```diff
@@ -0,0 +1,269 @@
...
+    import hvac
+    from hvac.api.secrets_engines.pki import DEFAULT_MOUNT_POINT
+    HAS_HVAC = True
+except ImportError:
+    HAS_HVAC = False
+
+
+def run_module():
+    if not HAS_HVAC:
+        raise AnsibleError("Please pip install hvac to use the hashi_vault modules.")
```
briantist, line 209: This will best be checked after instantiating `HashiVaultModule`, because the sanity checks try to look at the argspec to compare it to documentation and ensure they match. Also then you can use `module.fail_json()` since we don't have an exception class to raise.

plugins/modules/vault_pki_generate_certificate.py
```diff
@@ -0,0 +1,269 @@
...
+          If C(pem_bundle), the C(certificate) field will contain the private key and certificate, concatenated; if the issuing CA is not a Vault-derived
+          self-signed root, this will be included as well.
+      type: str
+      choices: [pem, der, pem_bundle]
+      default: pem
+    ip_sans:
+      description:
+        - Specifies requested IP Subject Alternative Names.
+        - Only valid if the role allows IP SANs (which is the default).
+      type: list
```
briantist, line 63: I think all of these `list` types should also include `elements: str`

plugins/modules/vault_pki_generate_certificate.py
```diff
@@ -0,0 +1,269 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+  module: vault_pki_generate_certificate
+  version_added: 2.3.0
+  author:
+    - Florent David <florent.david@gmail.com>
```
briantist, line 13: I believe we're removing/discouraging email addresses now in author fields and in copyright line too, you can replace email with your github ID in parentheses
```suggestion
    - Florent David (@Ripolin)
```

plugins/modules/vault_pki_generate_certificate.py
```diff
@@ -0,0 +1,269 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2022, Florent David <florent.david@gmail.com>
```
briantist, line 3:
```suggestion
# (c) 2022, Florent David (@Ripolin)
```

plugins/modules/vault_pki_generate_certificate.py
Ripolin: Fixed in commit d7c4dc98df2ae95e50ad4da21d3be935793f6668

plugins/modules/vault_pki_generate_certificate.py
Ripolin: Fixed in commit d7c4dc98df2ae95e50ad4da21d3be935793f6668

plugins/modules/vault_pki_generate_certificate.py
Ripolin: Fixed in commit d7c4dc98df2ae95e50ad4da21d3be935793f6668

plugins/modules/vault_pki_generate_certificate.py
Ripolin: Fixed in commit d7c4dc98df2ae95e50ad4da21d3be935793f6668

briantist:
Hey @Ripolin I'm hoping I'll have some time to dig into this further over the weekend, if you haven't already (I can't see the setting and haven't tried yet), can you check the box to enable maintainers to push to the PR? If you'd prefer I don't push and only suggest, let me know, that's fine but might mean longer turnaround.

Ripolin:
Hi @briantist. To me it's not a problem and box is already checked if you want to add some commits.

plugins/modules/vault_pki_generate_certificate.py
```diff
@@ -232,7 +232,6 @@ def run_module():
 
     if not HAS_HVAC:
         module.fail_json(msg=missing_required_lib('hvac'), exception=HVAC_IMPORT_ERROR)
-        return
```
Ripolin: Why remove this return ? To me module should be skip as quick as possible when import error is detected, that's the purpose of this return.

Ripolin:
@briantist Ok I will checks your modifications tomorrow and thx for your work.

briantist:
New changes since last comment:

- added unit tests (yay coverage!)
- added check mode support (with integration test for check mode)

I am not worried about the covecov "failure". There's a change here to exclude `conftest.py` files from coverage, but codecov doesn't seem to apply those changes until they are in `main`, and the "decrease" in coverage comes primarily from a generic fixture.

plugins/modules/vault_pki_generate_certificate.py
```diff
@@ -0,0 +1,282 @@
...
+
+    module.connection_options.process_connection_options()
+    client_args = module.connection_options.get_hvac_connection_options()
+    client = module.helper.get_vault_client(**client_args)
+
+    try:
+        module.authenticator.validate()
+        module.authenticator.authenticate(client)
+    except (NotImplementedError, HashiVaultValueError) as e:
+        module.fail_json(msg=to_native(e), exception=traceback.format_exc())
```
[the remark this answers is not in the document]
briantist: I am noticing that too. For now, I'd like to leave it in place, and I will consider a solution for this in the future and apply it retroactively.

plugins/modules/vault_pki_generate_certificate.py
```diff
@@ -0,0 +1,282 @@
...
+        ttl=dict(type='str', required=False, default=None),
+        format=dict(type='str', required=False, choices=['pem', 'der', 'pem_bundle'], default='pem'),
+        private_key_format=dict(type='str', required=False, choices=['der', 'pkcs8'], default='der'),
+        exclude_cn_from_sans=dict(type='bool', required=False, default=False),
+        engine_mount_point=dict(type='str', required=False)
+    )
+
+    module = HashiVaultModule(
+        argument_spec=argspec,
+        supports_check_mode=True
```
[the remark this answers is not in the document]
briantist: I have added the requisite lines for it to support check mode (see 263).

If you mean that it is not idempotent in the sense that it may perform a login, that is true and something to consider across the whole collection. I am starting to add such information to our docsite, you may review #211 if you'd like to see some of that!

---

For this module in particular, even though the status will always be `changed`, I believe that supporting check mode is more valuable than not. If we decline check mode support, Ansible will skip this module in check mode, and so it will show a skipped status.

That's good for safety, one of the two big reasons to use check mode, it's not so good for the other, which is to _report_ on what will change and what will not. By supporting check mode, we provide a `changed` status which is (at least slightly) more useful, in my opinion.

Ripolin:
LGTM

briantist (approved):

briantist:
Thanks for contributing this @Ripolin !

briantist:
@Ripolin [this is now released in version 2.3.0](https://github.com/ansible-collections/community.hashi_vault/releases/tag/2.3.0), thanks again!

51619742:
>

51619742:
@51619742

51619742:
- [ ]

PR #220, opened 2022-02-13
add k8s auth  to lookup hashi_vault - operator

chris93111:
##### SUMMARY
Add authentification k8s in lookup, for ansible operator (operator-sdk) use service account jwt token in login

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
lookup hashi_vault

```
- name: vault test jwt k8s
  hosts: localhost
  connection: localhost
  tasks:
    - name: try fetch secret with jwt SA
      debug:
        msg: "{{ lookup('community.hashi_vault.hashi_vault', 'secret=ansible/data/tower/credential/user1 auth_method=k8s mount_point=kubernetes-cluster1 role_id=awx url=https://vault:443')}}"

```

### DIFF
--- a/plugins/doc_fragments/auth.py
+++ b/plugins/doc_fragments/auth.py
@@ -27,6 +27,7 @@ class ModuleDocFragment(object):
           - jwt
           - cert
           - none
+          - k8s
         default: token
         type: str
       mount_point:
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -181,6 +181,18 @@ DOCUMENTATION = """
         - section: hashi_vault_collection
           key: token_validate
           version_added: 1.4.0
+    role:
+      ini:
+        - section: lookup_hashi_vault
+          key: role
+          deprecated:
+            why: collection-wide config section
+            version: 3.0.0
+            collection_name: community.hashi_vault
+            alternatives: use section [hashi_vault_collection]
+        - section: hashi_vault_collection
+          key: role
+          version_added: 2.3.0
     role_id:
       ini:
         - section: lookup_hashi_vault
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -181,18 +181,6 @@ DOCUMENTATION = """
         - section: hashi_vault_collection
           key: token_validate
           version_added: 1.4.0
-    role:
-      ini:
-        - section: lookup_hashi_vault
-          key: role
-          deprecated:
-            why: collection-wide config section
-            version: 3.0.0
-            collection_name: community.hashi_vault
-            alternatives: use section [hashi_vault_collection]
-        - section: hashi_vault_collection
-          key: role
-          version_added: 2.3.0
     role_id:
       ini:
         - section: lookup_hashi_vault
--- a/plugins/module_utils/_auth_method_k8s.py
+++ b/plugins/module_utils/_auth_method_k8s.py
@@ -20,13 +20,13 @@ class HashiVaultAuthMethodK8S(HashiVaultAuthMethodBase):
     '''HashiVault option group class for auth: k8s'''
 
     NAME = 'k8s'
-    OPTIONS = ['jwt', 'role', 'mount_point']
+    OPTIONS = ['jwt', 'role_id', 'mount_point']
 
     def __init__(self, option_adapter, warning_callback):
         super(HashiVaultAuthMethodK8S, self).__init__(option_adapter, warning_callback)
 
     def validate(self):
-        self.validate_by_required_fields('role')
+        self.validate_by_required_fields('role_id')
 
     def authenticate(self, client, use_token=True):
         params = self._options.get_filled_options(*self.OPTIONS)
@@ -38,6 +38,8 @@ class HashiVaultAuthMethodK8S(HashiVaultAuthMethodBase):
                 params['jwt'] = jwt
             except:
                 raise NotImplementedError("Can't read jwt in /var/run/secrets/kubernetes.io/serviceaccount/token")
+        params['role'] = params.pop('role_id')
+        
         try:
             response = client.auth_kubernetes(**params)
         except (NotImplementedError, AttributeError):
--- a/plugins/module_utils/_auth_method_k8s.py
+++ b/plugins/module_utils/_auth_method_k8s.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2021 FERREIRA Christophe (@chris93111)
+# Simplified BSD License (see licenses/simplified_bsd.txt or https://opensource.org/licenses/BSD-2-Clause)
+
+'''Python versions supported: all controller-side versions, all remote-side versions except 2.6'''
+
+# FOR INTERNAL COLLECTION USE ONLY
+# The interfaces in this file are meant for use within the community.hashi_vault collection
+# and may not remain stable to outside uses. Changes may be made in ANY release, even a bugfix release.
+# See also: https://github.com/ansible/community/issues/539#issuecomment-780839686
+# Please open an issue if you have questions about this.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+from ansible_collections.community.hashi_vault.plugins.module_utils._hashi_vault_common import HashiVaultAuthMethodBase
+
+
+class HashiVaultAuthMethodK8S(HashiVaultAuthMethodBase):
+    '''HashiVault option group class for auth: k8s'''
+
+    NAME = 'k8s'
+    OPTIONS = ['jwt', 'role', 'mount_point']
+
+    def __init__(self, option_adapter, warning_callback):
+        super(HashiVaultAuthMethodK8S, self).__init__(option_adapter, warning_callback)
+
+    def validate(self):
+        self.validate_by_required_fields('role')
+
+    def authenticate(self, client, use_token=True):
+        params = self._options.get_filled_options(*self.OPTIONS)
+        if not params['jwt']:
+            # Mode in cluster fetch jwt in pods
+            try:
+                f = open('/var/run/secrets/kubernetes.io/serviceaccount/token')
+                jwt = f.read()
+                params['jwt'] = jwt
+            except:
+                raise NotImplementedError("Can't read jwt in /var/run/secrets/kubernetes.io/serviceaccount/token")
+        try:
+            response = client.auth_kubernetes(**params)
+        except (NotImplementedError, AttributeError):
+            raise NotImplementedError("K8S authentication requires HVAC version 0.8.0 or higher.")
+
+        return response
--- a/plugins/module_utils/_auth_method_k8s.py
+++ b/plugins/module_utils/_auth_method_k8s.py
@@ -30,7 +30,7 @@ class HashiVaultAuthMethodK8S(HashiVaultAuthMethodBase):
 
     def authenticate(self, client, use_token=True):
         params = self._options.get_filled_options(*self.OPTIONS)
-        if not params['jwt']:
+        if not params.get('jwt'):
             # Mode in cluster fetch jwt in pods
             try:
                 f = open('/var/run/secrets/kubernetes.io/serviceaccount/token')
--- a/plugins/module_utils/_authenticator.py
+++ b/plugins/module_utils/_authenticator.py
@@ -57,7 +57,6 @@ class HashiVaultAuthenticator():
         aws_iam_server_id=dict(type='str'),
         cert_auth_private_key=dict(type='path', no_log=False),
         cert_auth_public_key=dict(type='path'),
-        role=dict(type='str'),
     )
 
     def __init__(self, option_adapter, warning_callback):
--- a/plugins/module_utils/_authenticator.py
+++ b/plugins/module_utils/_authenticator.py
@@ -22,6 +22,7 @@ from ansible_collections.community.hashi_vault.plugins.module_utils._auth_method
 from ansible_collections.community.hashi_vault.plugins.module_utils._auth_method_none import HashiVaultAuthMethodNone
 from ansible_collections.community.hashi_vault.plugins.module_utils._auth_method_token import HashiVaultAuthMethodToken
 from ansible_collections.community.hashi_vault.plugins.module_utils._auth_method_userpass import HashiVaultAuthMethodUserpass
+from ansible_collections.community.hashi_vault.plugins.module_utils._auth_method_k8s import HashiVaultAuthMethodK8S
 
 
 class HashiVaultAuthenticator():
@@ -36,6 +37,7 @@ class HashiVaultAuthenticator():
             'jwt',
             'cert',
             'none',
+            'k8s',
         ]),
         mount_point=dict(type='str'),
         token=dict(type='str', no_log=True, default=None),
@@ -55,6 +57,7 @@ class HashiVaultAuthenticator():
         aws_iam_server_id=dict(type='str'),
         cert_auth_private_key=dict(type='path', no_log=False),
         cert_auth_public_key=dict(type='path'),
+        role=dict(type='str'),
     )
 
     def __init__(self, option_adapter, warning_callback):
@@ -66,6 +69,7 @@ class HashiVaultAuthenticator():
             'aws_iam': HashiVaultAuthMethodAwsIam(option_adapter, warning_callback),
             'cert': HashiVaultAuthMethodCert(option_adapter, warning_callback),
             'jwt': HashiVaultAuthMethodJwt(option_adapter, warning_callback),
+            'k8s': HashiVaultAuthMethodK8S(option_adapter, warning_callback),
             'ldap': HashiVaultAuthMethodLdap(option_adapter, warning_callback),
             'none': HashiVaultAuthMethodNone(option_adapter, warning_callback),
             'token': HashiVaultAuthMethodToken(option_adapter, warning_callback),


plugins/module_utils/_auth_method_k8s.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2021 FERREIRA Christophe (@chris93111)
+# Simplified BSD License (see licenses/simplified_bsd.txt or https://opensource.org/licenses/BSD-2-Clause)
+
+'''Python versions supported: all controller-side versions, all remote-side versions except 2.6'''
```
briantist, line 5:
```suggestion
'''Python versions supported: >=3.6'''
```
This is a good reminder I need to update these in the other files.

plugins/module_utils/_auth_method_k8s.py
```diff
@@ -0,0 +1,48 @@
...
+__metaclass__ = type
+
+from ansible_collections.community.hashi_vault.plugins.module_utils._hashi_vault_common import HashiVaultAuthMethodBase
+
+
+class HashiVaultAuthMethodK8S(HashiVaultAuthMethodBase):
+    '''HashiVault option group class for auth: k8s'''
+
+    NAME = 'k8s'
+    OPTIONS = ['jwt', 'role_id', 'mount_point']
```
briantist, line 23: I'm going to recommend that we do  not re-use the `jwt` option name from JWT auth, and give this its own unique option, `kubernetes_token` or something like that.

plugins/doc_fragments/auth.py
```diff
@@ -27,6 +27,7 @@ class ModuleDocFragment(object):
           - jwt
           - cert
           - none
+          - k8s
```
briantist, line 30: I think we should give it a fuller name and call the method `kubernetes` as they do in Vault itself.
This will need a lot of renames but hopefully it's all simple find/replace.

plugins/module_utils/_auth_method_k8s.py
```diff
@@ -0,0 +1,48 @@
...
+        params = self._options.get_filled_options(*self.OPTIONS)
+        if not params.get('jwt'):
+            # Mode in cluster fetch jwt in pods
+            try:
+                f = open('/var/run/secrets/kubernetes.io/serviceaccount/token')
+                jwt = f.read()
+                params['jwt'] = jwt
+            except:
+                raise NotImplementedError("Can't read jwt in /var/run/secrets/kubernetes.io/serviceaccount/token")
+        params['role'] = params.pop('role_id')
```
briantist, line 41: This part should be handled in the `validate` method, similar to what's done for token auth.

The path to the fallback token should also be set in an option (or two, if it's advantageous to separate path and file), again similar to what token auth does, so that it can be configured and tested.

This is also another reason why the JWT for this method should have its own option, so that this fallback can be described properly and given dedicated fallback vars.

plugins/module_utils/_authenticator.py
```diff
@@ -22,6 +22,7 @@
 from ansible_collections.community.hashi_vault.plugins.module_utils._auth_method_none import HashiVaultAuthMethodNone
 from ansible_collections.community.hashi_vault.plugins.module_utils._auth_method_token import HashiVaultAuthMethodToken
 from ansible_collections.community.hashi_vault.plugins.module_utils._auth_method_userpass import HashiVaultAuthMethodUserpass
+from ansible_collections.community.hashi_vault.plugins.module_utils._auth_method_k8s import HashiVaultAuthMethodK8S
```
briantist, line 25: nit: alphabetize (see comment)

plugins/module_utils/_authenticator.py
```diff
@@ -66,6 +68,7 @@ def __init__(self, option_adapter, warning_callback):
             'aws_iam': HashiVaultAuthMethodAwsIam(option_adapter, warning_callback),
             'cert': HashiVaultAuthMethodCert(option_adapter, warning_callback),
             'jwt': HashiVaultAuthMethodJwt(option_adapter, warning_callback),
+            'k8s': HashiVaultAuthMethodK8S(option_adapter, warning_callback),
```
briantist, line 71: nit: alphabetize (see comment)

briantist (commented):
Hi @chris93111 ! Welcome and thanks for submitting!

I've put some inline comments in.

One of them suggestes changing the name of the auth method to `kubernetes`, since this is a more descriptive name, and it matches what the method is called in Vault's documentation.

Beyond that, the main thing missing is tests. Auth methods in this collection are shared by all plugins and modules (including ones that don't exist yet), so we really need them to be well tested. The other auth methods have unit tests and integration tests that cover both controller and target execution. There should be a lot to go on in the existing ones.

If you haven't seen it already, we do have a contributor guide with a section on adding auth methods:
https://docs.ansible.com/ansible/devel/collections/community/hashi_vault/docsite/contributor_guide.html#contributing-auth-methods
(suggestions for improving this guide are welcome too!)

I will also help out with testing and such where I can, but I can't guarantee when I'll be able to do so.

Thanks again and let me know if you have any questions.

chris93111:
@briantist can you please relaunch the CI and tell me if correction is ok ? they are error Node_error on last try

I can write test, but how i can mount minikube in the ci ? Tests can't work without

briantist:
> @briantist can you please relaunch the CI and tell me if correction is ok ? they are error Node_error on last try

Sure I've launched the CI for the latest commit but canceled the queued runs for the other commits. New contributors need approval to run on every push. I strongly recommend getting set up to run the tests locally. All the tests we run in CI are able to be run on a local machine, and there's instructions in the contributor guide, it will really make turnaround time a lot faster for you. If you have any trouble getting set up, let me know.

I don't have much left in me this weekend to take a deeper look at this yet, but I will as soon as I get more free time.

> I can write test, but how i can mount minikube in the ci ? Tests can't work without

Forgive me for being unfamiliar with minikube but from the looks of it, it's something we could possibly spin up as part of the integration tests. We already have Vault, tinyproxy, and MMock running, each in their own containers.

From what little I cold gather, minikube is not really designed to run _in_ a container. But we could launch it separately, as long as we ensure it works in both local and CI environments.

Looking around a bit, it seems like k3d (https://docs.ansible.com/ansible/devel/collections/community/hashi_vault/docsite/contributor_guide.html#mocking-integrations) might be better for this purpose? It looks to be designed to run in a container itself, so this would be even easier to run maybe.

To add a new piece of software like this, you'll want to look at the [localenv developer guide](https://docs.ansible.com/ansible/devel/collections/community/hashi_vault/docsite/localenv_developer_guide.html), and at the [current docker-based localenv](https://github.com/ansible-collections/community.hashi_vault/tree/main/tests/integration/targets/setup_localenv_docker) that we use to spin up the current set of containers.

---

But there is an easier option too, we can [mock the integration with MMock](https://github.com/ansible-collections/community.hashi_vault/tree/main/tests/integration/targets/setup_localenv_docker). We already do this for other things that weren't so easy to set up (AWS auth and LDAP auth). You can see the mmock configs here: 
https://github.com/ansible-collections/community.hashi_vault/tree/main/tests/integration/targets/setup_localenv_docker/templates/mmock

And if you run this locally, you'll be able to connect to MMock's console port in your web browser and get a real-time view of requests which can help tune it.

I can help with that if you have the server response from running the real command against a Vault server (with everything sensitive redacted/replaced).

The basic idea is that the integration tests point at MMock, and MMock is set up to proxy those requests to the test Vault server, unless there's a config for the request that matches a given path (or other criteria), and for those requests, we can return a mock response.

---

Integration tests aside, you can start writing unit tests right away, which can cover a lot of corner cases more easily than integration tests can, and they will not/can not talk to any external services.

Grabbing a real response as described above is useful for unit tests as well as MMock, you can see the other ones we have here:
https://github.com/ansible-collections/community.hashi_vault/tree/main/tests/unit/fixtures

And we use these as mock returns in the unit tests as well.

---

Hopefully there's plenty here that you can get started on, even if I can't put a lot of deep time into this just yet, I will try my best to answer questions and keep you moving along!

plugins/doc_fragments/auth.py
```diff
@@ -72,6 +73,13 @@ class ModuleDocFragment(object):
       jwt:
         description: The JSON Web Token (JWT) to use for JWT authentication to Vault.
         type: str
+      kubernetes_token:
+        description: The Kubernetes Token (JWT) to use for Kubernetes authentication to Vault.
+        type: str
+      kubernetes_token_path:
+        description: If no kubernetes_token is specified, will try to read the token from this path.
+        default: '/var/run/secrets/kubernetes.io/serviceaccount/token'
+        type: str
```
briantist, line 82: When we're adding a new option, we can add `version_added` at the option level to mark when it was added, and that applies to env/ini/vars too (those have separate fields in case they were added to an option later).

I'm about to release 2.3.0 as there's some new stuff merged, so this will be slated for 2.4.0 for now. Not to worry though, I don't have a set release schedule for this collection, typically I do a release when there's something new.

```suggestion
      kubernetes_token:
        description: The Kubernetes Token (JWT) to use for Kubernetes authentication to Vault.
        type: str
        version_added: 2.4.0
      kubernetes_token_path:
        description: If no kubernetes_token is specified, will try to read the token from this path.
        default: '/var/run/secrets/kubernetes.io/serviceaccount/token'
        type: str
        version_added: 2.4.0
```

plugins/doc_fragments/auth.py
```diff
@@ -245,4 +253,22 @@ class ModuleDocFragment(object):
...
+        env:
+          - name: ANSIBLE_HASHI_VAULT_KUBERNETES_TOKEN_PATH
+            version_added: 2.3.0
+        ini:
+          - section: hashi_vault_collection
+            key: kubernetes_token_path
+            version_added: 2.3.0
+        vars:
+          - name: ansible_hashi_vault_kubernetes_token_path
+            version_added: 2.3.0
```
briantist, line 273: We don't need `version_added` within these unless they were added after the option itself, so we can just put that annotation right on the whole option (see previous comment)

```suggestion
      kubernetes_token:
        env:
          - name: ANSIBLE_HASHI_VAULT_KUBERNETES_TOKEN
        vars:
          - name: ansible_hashi_vault_kubernetes_token
      kubernetes_token_path:
        env:
          - name: ANSIBLE_HASHI_VAULT_KUBERNETES_TOKEN_PATH
        ini:
          - section: hashi_vault_collection
            key: kubernetes_token_path
        vars:
          - name: ansible_hashi_vault_kubernetes_token_path
```

plugins/lookup/hashi_vault.py
```diff
@@ -205,6 +205,30 @@
...
+        - section: lookup_hashi_vault
+          key: kubernetes_token_path
+          deprecated:
+            why: collection-wide config section
+            version: 3.0.0
+            collection_name: community.hashi_vault
+            alternatives: use section [hashi_vault_collection]
+        - section: hashi_vault_collection
+          key: token_path
+          version_added: 2.3.0
```
briantist, line 231: Thank you for thinking about these!
We actually only needed to do that for options that existed before the deprecation was made, to give people time to migrate.
Any new options added since the change was announced will only exist in the new INI section, so we won't need this here.

```suggestion
```

plugins/module_utils/_auth_method_k8s.py
```diff
@@ -0,0 +1,53 @@
...
+            raise HashiVaultValueError(self._options.get_option('kubernetes_token')+self._options.get_option_default('kubernetes_token_path')+"No Kubernetes Token specified or discovered.")
+
+    def authenticate(self, client, use_token=True):
+        origin_params = self._options.get_filled_options(*self.OPTIONS)
+        params = {"role": origin_params.get('role_id'),
+                  "jwt": origin_params.get('kubernetes_token'),
+                  "mount_point": origin_params.get('mount_point')}
+
+        try:
+            response = client.auth_kubernetes(**params)
```
briantist, line 49: This method is deprecated and slated for removal in HVAC 1.0.0 (which is the next expected version). The new method is `client.auth.kubernetes.login` (see https://hvac.readthedocs.io/en/stable/source/hvac_api_auth_methods.html#hvac.api.auth_methods.Kubernetes.login).

I realize the example shown in HVAC's docs still uses the old method.

What we've done typically for these things is that we try the new method and then fallback to the deprecated method with a warning, that way we can still support older hvac versions for now.

For example, here's how it was done in the LDAP auth method:
https://github.com/ansible-collections/community.hashi_vault/blob/fdb0ee07d7057319b4f20e3174f2421e335c9e30/plugins/module_utils/_auth_method_ldap.py#L31-L37

Also this call is not making use of the `authenticate` method's `use_token` parameter.

chris93111: ok

chris93111:
Hi @briantist can you please review this before i try to work with CI

plugins/module_utils/_auth_method_k8s.py
```diff
@@ -37,7 +38,8 @@ def validate(self):
                     self._options.set_option('kubernetes_token', token_file.read().strip())
 
         if self._options.get_option('kubernetes_token') is None:
-            raise HashiVaultValueError(self._options.get_option('kubernetes_token')+self._options.get_option_default('kubernetes_token_path')+"No Kubernetes Token specified or discovered.")
+            raise HashiVaultValueError(self._options.get_option_default('kubernetes_token_path') + 
+                                       " No Kubernetes Token specified or discovered.")
```
briantist, line 42: Formatting preference, and also changed to `get_option` because this option should never be missing (it has a default value right?).

```suggestion
            raise HashiVaultValueError(
                self._options.get_option('kubernetes_token_path') + 
                " No Kubernetes Token specified or discovered."
            )
```

plugins/doc_fragments/auth.py
```diff
@@ -27,6 +27,7 @@ class ModuleDocFragment(object):
           - jwt
           - cert
           - none
+          - kubernetes
```
briantist, line 30: Small thing and I don't know if it even makes sense, but let's move this above `none`, for some reason I like `none` being at the end.

Also please add a line after line 19, to describe when this method was added (see the other options there as well).

plugins/module_utils/_authenticator.py
```diff
@@ -36,6 +37,7 @@ class HashiVaultAuthenticator():
             'jwt',
             'cert',
             'none',
+            'kubernetes',
```
briantist, line 40: When changing the list in `doc_fragments/auth.py`, please change this list to match.

briantist (commented):
some very small changes suggested

briantist:
A change was made in core a few hours ago that will cause sanity test failures (unrelated to your PR). I have a PR up to address those changes:
- https://github.com/ansible-collections/community.hashi_vault/pull/226

When that merges, you'll have to rebase to ensure sanity runs.

plugins/module_utils/_auth_method_k8s.py
chris93111: yes the defaut path of kubernetes token in pods

chris93111:
@briantist i thinks it's ok ?

briantist:
Looks like you've got one sanity nit in there, I recommend setting your editor to trim trailing whitespace, insert final newline (single), and trim final newlines (these are roughly the names of the settings in vscode, not sure what they are called in other editors).

Also again, running tests locally will really help you find these things quickly... and will be invaluable when you start adding the integration and unit tests.

If there's any way I can help you get your local setup going, let me know!

plugins/doc_fragments/auth.py
```diff
@@ -72,6 +73,15 @@ class ModuleDocFragment(object):
       jwt:
         description: The JSON Web Token (JWT) to use for JWT authentication to Vault.
         type: str
+      kubernetes_token:
+        description: The Kubernetes Token (JWT) to use for Kubernetes authentication to Vault.
+        type: str
+        version_added: 2.4.0
```
briantist, line 79:
```suggestion
        version_added: 2.5.0
```

plugins/doc_fragments/auth.py
```diff
@@ -72,6 +73,15 @@ class ModuleDocFragment(object):
...
+      kubernetes_token_path:
+        description: If no kubernetes_token is specified, will try to read the token from this path.
+        default: '/var/run/secrets/kubernetes.io/serviceaccount/token'
+        type: str
+        version_added: 2.4.0
```
briantist, line 84:
```suggestion
        version_added: 2.5.0
```

briantist (commented):
bump `version_added`

briantist:
Hi @chris93111 , are you still interested in pursuing this? 
I wanted to let you know that [version 3.0.0](https://github.com/ansible-collections/community.hashi_vault/milestone/15) of the collection will be released in the coming weeks, with some breaking changes. Support for Ansible 2.9 and 2.10 will be dropped and won't be tested against anymore either.

If you'd like to get this into a release that still supports 2.9/2.10, it would be best to get this moving along for the best chance of that. I will need to get 3.0.0 out in time for the deadline for inclusion into the Ansible 6 community package, so I don't really have any leeway when to release that.

If you're not concerned about that then no worries and no rush, just wanted to give you a heads-up.

chris93111:
hello @briantist 
Yes of course! I need to look at the important changes brought by the non support of 2.9 and 2.10
I also have to look at the CI

briantist:
> hello @briantist Yes of course! I need to look at the important changes brought by the non support of 2.9 and 2.10 I also have to look at the CI

Great! There's nothing very important you need to know about removal of 2.9/2.10 support for your implementation, I only mention it in case you need this collection with one of those versions, or you would like to get kubernetes auth released in a  version of this collection that users of those versions can trust for use.

I don't actually expect the first release of 3.0.0 to be broken in 2.9 or 2.10, but since we will stop testing against those versions, any new change in 3.x.x may break compatibility, so I expect anyone using 2.9 or 2.10 to pin to `community.hashi_vault < 3.0.0`.

---

Getting the unit tests and especially integration tests set up for this is absolutely going to be the most time consuming part of getting this ready to be merged. I believe it is entirely possible to do, but the timeline will depend on mostly on how much time you can devote to it, because I am very overloaded right now. 

I can be more effective now by helping your efforts and steering you in the right direction, rather than trying to implement myself, I think.

In my opinion, the fastest way to get this to a point where we can merge, is to use the MMock method in integration tests. That could give us reasonable confidence, and will not require setting up kubernetes. We can then look up at real kubernetes of some kind as a new PR.

briantist:
@chris93111 I'm not sure what happened with https://github.com/ansible-collections/community.hashi_vault/pull/220/commits/d49061d98741c7d03453797e17285bf51e2321b5 but the test results don't look right; it seems like perhaps some changes from `main` are missing.

PR #221, opened 2022-02-14
codecov - try different path format

briantist:
##### SUMMARY

I think I've been using the wrong paths for the ignore section, I was trying to start them with `**/` but since the paths started in the root, I think the path fixing was trimming off the beginnings (like it was supposed to), and so the `/` part was causing them not to match... trying to see if this works.

##### ISSUE TYPE

- CI Pull Request

##### COMPONENT NAME

n/a

##### ADDITIONAL INFORMATION

n/a

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,7 +1,7 @@
 ---
 ignore:
-  - "**/tests/unit/compat/*"
-  - "**/tests/unit/**/conftest.py"
+  - tests/unit/compat/*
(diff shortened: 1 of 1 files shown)


briantist:
Confirmed: changing this file alone raised coverage, without any other files being changed.
This also answers the question of whether codecov uses the yml file in the diff in calculating diff coverage, and it seems that it does!

PR #235, opened 2022-03-26
Docs/strict and lenient

briantist:
##### SUMMARY
Builds on the changes in https://github.com/ansible-community/github-docs-build/pull/27 .

* Set the push build to the strictest set of option
* On PR, set the PR build to the most lenient set of options; this gives us the best possibility of a publishable build, because a partial build of the docsite is more useful than none at all
* On PR, add an additional docs build (no comparison to base) with the most strict settings; this serves as the quality/validator for the PR, so even if the lenient build for the PR publish passes, this one will catch other errors and "fail" the PR

As with all `pull_request_target` workflow changes, the tests run in this PR do not test the changes to the workflow.

Test PR is needed against this branch to test functionality.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

docs-build

### DIFF
--- a/.github/workflows/docs-push.yml
+++ b/.github/workflows/docs-push.yml
@@ -18,6 +18,8 @@ jobs:
     with:
       ansible-ref: devel
       init-dest-dir: docs/preview
+      init-lenient: false
+      init-fail-on-error: true
 
   publish-docs:
     # for now we won't run this on forks
--- a/.github/workflows/docs-push.yml
+++ b/.github/workflows/docs-push.yml
@@ -10,23 +10,33 @@ on:
     - cron: '0 13 * * *'
 
 jobs:
+  validate-docs:
+    permissions:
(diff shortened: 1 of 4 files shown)


briantist:
I forgot that these options are for initialization only, and this collection skips init, so we can't use these.

Will have to reconsider how/whether to do this alongside having the committed init'd build env.

The reason we commit the env in this collection is:
* the slightly customize the main RST file (not terribly important)
* to give contributors an easier way to build docs locally without having to init themselves

A possible way to address this may be to further modify `build.sh` to take parameters.

To do that, would require options in the docs build workflows and build action. This might be a good idea anyway, I'm thinking of just making it a generic "extra args" type of thing that gets passed directly to the build script.

briantist:
Also needs: https://github.com/ansible-community/antsibull/pull/409

PR #246, opened 2022-04-23
add content for KV get operations

### LINKED ISSUE #9: hashi_vault should support secret v2 path (without adding /data/)
##### SUMMARY

The `hashi_vault` lookup currently needs to use an API path to kv_v2 secrets. It would be nice to drop this requirement, like terraform do.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

hashi_vault

##### ADDITIONAL INFORMATION

The current call:
```yaml
- name: get secret (currently)
  debug:
    msg: "{{ lookup('hashi_vault', 'engine/path/data/secret/path`) }}"
```

The wanted call:
```yaml
- name: get secret (wanted)
  debug:
    msg: "{{ lookup('hashi_vault', 'engine/path/secret/path`) }}"
```

In the terraform provider, they [get the engine path](https://github.com/terraform-providers/terraform-provider-vault/blob/61bd17181d12d43433be914cb015dc2d169fd574/vault/kv_helpers.go#L88) from a call to `/v1/sys/internal/ui/mounts/{{ FULL_SECRET_PATH }}`.

### LINKED ISSUE #251: several modules are missing from the action group
##### SUMMARY

I keep forgetting the action_group exists when adding new modules.

Missing:
- `vault_pki_generate_certificate`
- `vault_write`
- `vault_token_create`

##### ISSUE TYPE
- Bug report

briantist:
##### SUMMARY
Adding modules and lookups for KV get.  

I started with a single `vault_kv_get` that would take a `kv_version` option to decide which KV version we're dealing with, but ultimately decided on separate modules/plugins for kv1 and kv2: both the inputs (options) and outputs (return data) are different, so trying to mix them doesn't make a lot of sense.

I have mulled over the possibility of a generic `vault_kv_get` _in addition_ that could A) take a `kv_version`, B) read the mount point info from Vault to determine which KV version it is (extra round trip + requires permissions), or C) both (like `kv_version` could be `1`, `2`, or `'auto'`). I am still not so sure this "generic" one is a good idea. I won't be implementing it in this PR, but if there's demand we'll have a public discussion about it.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

vault_kv1_get (lookup & module)
vault_kv2_get (lookup & module)

##### ADDITIONAL INFORMATION

- https://github.com/ansible-collections/community.hashi_vault/discussions/229

- Resolves #9 
- Fixes #251


felixfontein (approved):
I only glanced over it (and ignored the tests :) ), but it looks good to me!

briantist:
thanks for taking a look @felixfontein !

plugins/lookup/vault_kv2_get.py
```diff
@@ -0,0 +1,230 @@
...
+        data:
+          data:
+            Key1: value1
+            Key2: value2
+          metadata:
+            created_time: "2022-04-21T15:56:58.8525402Z"
+            custom_metadata": null
+            deletion_time": ""
+            destroyed": false
+            version": 2
```
lowlydba, line 128: Super late, but are these missing opening quotes?

felixfontein: I think it's more that they have closing quotes that shouldn't be there :-)

This is actually valid YAML (just checked), you get a dictionary where the keys end with `"`. Which is likely not what the API returns.

briantist: yep, that was not intentional.. thanks @lowlydba ! 
I'll fix it when I fix #258 , before this stuff is released.

briantist: Fixed in https://github.com/ansible-collections/community.hashi_vault/pull/264 if you have a moment @lowlydba , thank you!!

PR #264, opened 2022-05-01
KV get fixups

### LINKED ISSUE #258: The `vault_kv*_get` content uses `backend_mount_point` but `vault_pki_generate_certificate` uses `engine_mount_point`
##### SUMMARY

The new content in #246 uses `backend_mount_point` but `vault_pki_generate_certificate` uses `engine_mount_point`.

These should use the same. I will change the new ones, and the doc fragment, to match `engine_mount_point` because:
* it is a better description (technically auth plugins are also ("auth") backends)
* it's already in use
* the new content is not released yet, so this is not a breaking change

##### ISSUE TYPE
- Feature Idea

briantist:
##### SUMMARY

Fixes #258 
Fixes https://github.com/ansible-collections/community.hashi_vault/pull/246#discussion_r862379062

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

vault_kv1_get
vault_kv2_get
vault_pki_generate_certificate

##### ADDITIONAL INFORMATION

The new plugins are not released yet, so there is no breaking change or changelog frragment needed.

### DIFF
--- a/plugins/doc_fragments/engine_mount.py
+++ b/plugins/doc_fragments/engine_mount.py
@@ -11,14 +11,14 @@ class ModuleDocFragment(object):
 
     DOCUMENTATION = r'''
 options:
-  backend_mount_point:
+  engine_mount_point:
     description: The path where the secret backend is mounted.
     type: str
 '''
 
     PLUGINS = r'''
 options:
-  backend_mount_point:
+  engine_mount_point:
     vars:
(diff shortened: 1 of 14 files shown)


felixfontein (approved):

briantist:
@felixfontein thank you for reviewing!

PR #265, opened 2022-05-08
Content/lookup config

### LINKED ISSUE #250: lookup plugin for retrieving collection config via plugin sources
The plugins in the collection take values from vars, env vars, and INI config. it's a very convenient place to set wide defaults like the Vault URL, but there's no single straightforward way to share those values with modules, or put them in module_defaults.

A lookup plugin (being a plugin) can access those values and return them, which can be used to populate the values, something like:

```yaml
- community.hashi_vault.vault_read:
    url: "{{ lookup('community.hashi_vault.vault_plugin_config', 'url') }}"
```

or 

```yaml
module_defaults:
  group/community.hashi_vault.vault: "{{ lookup('community.hashi_vault.vault_plugin_config', 'url', 'auth_method', 'retries') }}"
```

### Discussed in https://github.com/ansible-collections/community.hashi_vault/discussions/242

<div type='discussions-op-text'>

<sup>Originally posted by **briantist** October 26, 2021</sup>
##### SUMMARY
We have very centralized options, and in plugins, we have built-in capability to set those options via environment variables, INI (`ansible.cfg`) settings, Ansible variables, etc. across all of the controller-side plugins.

Modules however, do not use any of those values. This is at least partially intentional on the part of Ansible: modules execute on a _target_ host, not on the controller.

So we have only the `module_defaults:` keyword to be able to set defaults for these, and don't currently have a way to (re)infer the values from the same place as plugins (env vars can be read explicitly when setting `module_defaults` via the `env` lookup, ansible vars can just be used directly, but there's no way to get the INI config values). There is also no "ansible-wide" place to configure such defaults, so at minimum a `module_defaults` section is needed per play[book].

We also must consider again that using controller-side values may not be correct. This is especially important for sensitive values, like using the env vars and sink files for tokens.

So this issue is an acknowledgement of the current situation and a place to discuss various ways of addressing it.

We should consider utility, ease of use, intuitiveness in solutions, but also ensure we're encouraging good practices and working _within_ the Ansible ecosystem and its standards and practices, rather than against.</div>

briantist:
##### SUMMARY

Resolves #250

Adds a lookup plugin that can retrieve plugin configuration, for use in calling modules or setting `module_defaults`

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

vault_ansible_settings

##### ADDITIONAL INFORMATION

https://github.com/ansible-collections/community.hashi_vault/discussions/242


felixfontein (approved):

briantist:
thanks for reviewing this @felixfontein !

Issue #248, opened 2022-04-24
The default value of `token_validate` will change from `True` to `False` in version `4.0.0`


briantist:
##### SUMMARY

As a holdover from the original `hashi_vault` lookup, token auth does a `lookup-self` on the given token as a way to try to determine if the token is valid.

Later an option `token_validate` was added to control this (so that it could be turned off), because in some cases, tokens don't have permission to perform that operation.

But its default was set to `true` to match previous behavior.

However, this lookup is a largely unnecessary extra round-trip for everything that uses token auth, that's on by default.

The only real advantage of doing this lookup is that we can provide a more specific error message if this fails. Without this lookup, the token will attempt to be used for whatever operation is being performed, and if it's not valid, the failure returned will depend on the operation somewhat, and so it may look like a different error (permission denied for example).

The disadvantages of doing this "validation" are:
* It's not actually validating the validity of the token
* It does not determine whether the token has the permission needed to perform the operation it needs to perform
* It could prevent unauthenticated endpoints from working if an old inferred token is read (from the sink file for example), unless `none` auth is explicitly chosen (arguably, you should be choosing `none` if you are hitting an unauthenticated endpoint)
* It's an extra round-trip

---

The only other area where this matters is in the `vault_login` lookup/module. Using `token` auth in `vault_login` returns the token itself (there is no login operation to perform), but it has to use `lookup-self` to get any additional information about the token, otherwise it will only return the token value (in the familiar structure).

So I have two options for `vault_login`:
1. Let the default change (keep inheriting from the global definition)
1. Change the default globally, but override the default in `vault_login` to be `true`.

~I am leaning toward option 1; the other information returned is the information least likely to be used, the option can be set explicitly, and there will be sufficient announcement of the change.~

After some testing, I will go with option 2. The issue is larger than just "extra" information, it's that with `vault_login` there is no "actual" operation to perform with a "token" login that will fail. So if you call `vault_login` with "token" auth and a bad/expired token, it will just return that (it will "succeed"). So I think validation defaulting to true for `vault_login` only makes sense.

##### ISSUE TYPE
- Feature Idea

PR #267, opened 2022-05-08
implement `token_validate` default change

briantist:
##### SUMMARY

Related: #248

Announce that the default value for `token_validate` is changing in 4.0.0.

Will repeat this PR in the release for 3.0.0 so that it's announced in the porting guide for 4.0.0 too.

### Which changelog section?

I used `deprecated_features` but I don't think it's quite right. Nothing is being deprecated, just changed.

I'm trying to consider if `breaking_changes` is more appropriate, either now for the announcements, in 4.0.0 for the actual change, or even both?

`major_features` feels too major. For the actual change, `removed_features` doesn't really make sense.

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/plugins/doc_fragments/auth.py
+++ b/plugins/doc_fragments/auth.py
@@ -52,8 +52,8 @@ class ModuleDocFragment(object):
         description:
           - For token auth, will perform a C(lookup-self) operation to determine the token's validity before using it.
           - Disable if your token does not have the C(lookup-self) capability.
-          - The default value is C(True).
-          - The default value will change to C(False) in version 4.0.0.
+          - The default value is C(true).
+          - The default value will change to C(false) in version 4.0.0.
         type: bool
         version_added: 0.2.0
       username:
--- a/plugins/doc_fragments/auth.py
+++ b/plugins/doc_fragments/auth.py
@@ -52,8 +52,9 @@ class ModuleDocFragment(object):
         description:
           - For token auth, will perform a C(lookup-self) operation to determine the token's validity before using it.
           - Disable if your token does not have the C(lookup-self) capability.
+          - The default value is C(True).
+          - The default value will change to C(False) in version 4.0.0.
         type: bool
-        default: true
         version_added: 0.2.0
       username:
         description: Authentication user name.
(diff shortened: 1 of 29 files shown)


felixfontein:
> I used `deprecated_features` but I don't think it's quite right. Nothing is being deprecated, just changed.

The old behavior of using the old default value is deprecated. So IMO this is ok :)

I would also add some code changes though so that users are informed that if they don't specify the value explicitly they will have a breaking change.

briantist:
> > I used `deprecated_features` but I don't think it's quite right. Nothing is being deprecated, just changed.
> 
> The old behavior of using the old default value is deprecated. So IMO this is ok :)

Thanks! Ok I'll leave it.


> I would also add some code changes though so that users are informed that if they don't specify the value explicitly they will have a breaking change.

I was thinking about this. If I warn all that time that will be pretty annoying. 

Unfortunately, I can't tell easily when the value was gotten from defaults, unless I use the same type of config manager code I used in the settings lookup.

Actually, just running that lookup from within other plugins would be an easy way to warn selectively without (re)writing a bunch of code.. it would only work in plugins though. 

In modules, I guess I could change the true default to None and then set it after I know.. i think sanity would not like it unless i change the docs too which is not great.

What do you think @felixfontein ?

felixfontein:
I did reply on IRC, for sake of completeness I'll repost it here:

> the 'common way' is to remove the default and in case the value is None, issue a deprecation warning and manually set the default value
> you can't use `default: xxx` in DOCUMENTATION though for modules (due to sanity tests), but you can't do that for plugins either, so the common thing is to mention this in in the regular description of the option

plugins/doc_fragments/auth.py
```diff
@@ -52,8 +52,9 @@ class ModuleDocFragment(object):
         description:
           - For token auth, will perform a C(lookup-self) operation to determine the token's validity before using it.
           - Disable if your token does not have the C(lookup-self) capability.
+          - The default value is C(True).
```
felixfontein, line 55: In documentation Ansible usually uses the lower-case versions.
```suggestion
          - The default value is C(true).
```

felixfontein (approved):

briantist:
Thank you for reviewing @felixfontein !

PR #273, opened 2022-05-13
announce `token_validate` default change in 3.0.0

briantist:
##### SUMMARY

Related:
- #248 

This was announced in 2.5.0, this PR is adding the announcement in 3.0.0.

##### ISSUE TYPE

- deprecation/removal

### DIFF
--- a/changelogs/fragments/248-token_validate-change-default.yml
+++ b/changelogs/fragments/248-token_validate-change-default.yml
@@ -1,0 +1,3 @@
+---
+deprecated_features:
+  - token_validate options - the shared auth option ``token_validate`` will change its default from ``True`` to ``False`` in community.hashi_vault version 4.0.0. The ``vault_login`` lookup and module will keep the default value of ``True`` (https://github.com/ansible-collections/community.hashi_vault/issues/248).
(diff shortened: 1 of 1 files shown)


changelogs/fragments/248-token_validate-change-default.yml
```diff
@@ -0,0 +1,3 @@
+---
+deprecated_features:
+  - token_validate options - the shared auth option ``token_validate`` will change its default from ``True`` to ``False`` in community.hashi_vault version 4.0.0. The ``vault_login`` lookup and module will keep the default value of ``True`` (https://github.com/ansible-collections/community.hashi_vault/issues/248).
```
felixfontein, line 3: In Ansible docs, we usually use lower-case names for the booleans. But if you use the upper-case version everywhere, feel free to keep this as-is.

briantist, line 3:
```suggestion
  - token_validate options - the shared auth option ``token_validate`` will change its default from ``true`` to ``false`` in community.hashi_vault version 4.0.0. The ``vault_login`` lookup and module will keep the default value of ``true`` (https://github.com/ansible-collections/community.hashi_vault/issues/248).
```

felixfontein (approved):

Issue #275, opened 2022-06-02
Simplified BSD License is referenced but not included


briantist:
##### SUMMARY

Some module_utils and such use the Simplified BSD License but the file itself is not included in the collection. 
Partially addressed in #280, the rest in #282 .

See also https://github.com/ansible-collections/community.general/pull/4759 for example.

PR #280, opened 2022-07-16
License fixes (part 1)

### LINKED ISSUE #276: Collection violates Ansible Community Package licensing guidelines
https://github.com/ansible-collections/community.hashi_vault/blob/8c0fc19d125d4e979888be2c11e8a6d3c67de94d/plugins/plugin_utils/_hashi_vault_lookup_base.py#L2-L3

According to the Ansible Community Package licensing guidelines: "All other code [i.e. not modules or module_utils] must be under the [GPL-3.0-or-later](https://www.gnu.org/licenses/gpl-3.0-standalone.html). These plugins are run inside of the Ansible controller process which is licensed under the GPLv3+ and often must import code from the controller. For these reasons, the GPLv3+ must be used."

See https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#id7 for more information. I'm not sure what is the procedure to rectify this, but I noticed it while preforming a license review for the Fedora ansible package and figured I'd point it out. It looks like @briantist is the only copyright owner/commiter, so I guess they'd have to relicense the file in question.

briantist:
##### SUMMARY

Fixes #276
Partially addresses #275
Related: https://github.com/ansible-community/community-topics/issues/112

For #275, the missing piece is updating all the headers.

I will update all the headers in a separate PR, so that I can add the resulting single commit to `.git-blame-ignore-revs` (see [ignore commits in the blame view](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)).

##### ISSUE TYPE

- Bugfix Pull Request


plugins/plugin_utils/_hashi_vault_lookup_base.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright (c) 2021 Brian Scholer (@briantist)
-# Simplified BSD License (see licenses/simplified_bsd.txt or https://opensource.org/licenses/BSD-2-Clause)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 3: The GPL license file is called `LICENSE`, not `COPYING`, in this collection. (Might be wrong in more files?)

briantist: good catch thanks, I think the other files are correct, and that I just copied this from the wrong place, but I will check on that. Anyway if I implement the move to a directory, I'll also have to revisit all the headers.

briantist: Actually the other files are all wrong lol

briantist: This will be addressed in part 2:
- #282

felixfontein: I think this is a quite common mistake, many collections have that problem since the "big split" for Ansible 2.10. The collection template and the collection repos created back then were using `LICENSE` for the filename, while the comments in all modules/plugins migrated from ansible/ansible were referencing it as `COPYING`. I renamed `LICENSE` to `COPYING` in the collections I was working on (like community.general) to avoid changing all the headers, other collections adjust the headers, or didn't change anything ;-)

LICENSES/GPL-3.0-or-later.txt
```diff
@@ -0,0 +1 @@
+../LICENSE
```
felixfontein, line 1: I really wish the GitHub UI would make it easier to see whether this file is a symlink or not. (It does show in the file tree on the left, but I don't think it can be seen from the diff list on the right...)

There were multiple cases of PRs in c.g for new modules where authors actually added a text file with this content instead of a symlink... When the tree on the left didn't yet exist it was only possible to distinguish this from the errors the sanity tests emitted ;-)

LICENSES/simplified_bsd.txt
```diff
@@ -0,0 +1,7 @@
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
```
felixfontein, line 1: The filename should be `BSD-2-Clause.txt`.

briantist: ah ok, I had based it on this (earlier) PR 😅 : https://github.com/ansible-collections/community.general/pull/4759/files#diff-cf536a04b1dc7d11785cb3d01164cbada12d6613c885f58d808826274275a883

changelogs/fragments/licensing.yml
```diff
@@ -0,0 +1,4 @@
+---
+bugfixes:
+  - Use the correct GPL license for plugin_utils (https://github.com/ansible-collections/community.hashi_vault/issues/276).
+  - Add missing ``simplified_bsd.txt`` file for BSD licensed content (https://github.com/ansible-collections/community.hashi_vault/issues/275).
```
felixfontein, line 4: It's no longer called this :)

briantist: ah right, I also forgot to add a changelog entry for #282 (since I don't want to put it in the ignored commit), updated in https://github.com/ansible-collections/community.hashi_vault/pull/280/commits/ce2917b23ca832c4ebe2d76e1fb95a13450e73d3

felixfontein (approved):

briantist:
Thank you for reviewing @felixfontein !

Issue #279, opened 2022-07-14
The default for `vault_kv2_get` lookup's `engine_mount_point` will change to `secret` in version `4.0.0`


briantist:
##### SUMMARY

`engine_mount_point` is supposed to default to `secret` in `vault_k2_get`, and it does in the module, but in the lookup it was accidentally set to `kv` (which is the default for `kv1`).

Although this was not intentional, changing it now is a breaking change, so it will be done in version 4 of the collection.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

vault_kv2_get

##### COLLECTION VERSION

```paste below
3.0.0
```

PR #281, opened 2022-07-16
Deprecate incorrect `vault_kv2_get` lookup default

briantist:
##### SUMMARY

Implements the deprecation for #279 in advance of the change in 4.0.0.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

vault_kv2_get

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -2,7 +2,7 @@
 
 namespace: community
 name: hashi_vault
-version: 3.1.0
+version: 3.0.1
 readme: README.md
 authors:
   - Julie Davila (@juliedavila) <julie(at)davila.io>
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -2,7 +2,7 @@
 
 namespace: community
 name: hashi_vault
-version: 3.0.1
+version: 3.1.0
 readme: README.md
 authors:
   - Julie Davila (@juliedavila) <julie(at)davila.io>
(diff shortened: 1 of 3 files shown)


galaxy.yml
```diff
@@ -2,7 +2,7 @@
 
 namespace: community
 name: hashi_vault
-version: 3.1.0
+version: 3.0.1
```
felixfontein, line 5: Deprecations must appear in minor releases.

felixfontein (commented):
Besides this, LGTM.

galaxy.yml
briantist: oh interesting, I don't think I ever realized that, thanks!

felixfontein: See the semver spec, point 7: https://semver.org/

> Minor version Y (x.Y.z | x > 0) [...] MUST be incremented if any public API functionality is marked as deprecated.

briantist: I see, I guess I sort of felt like we're using the deprecation system sometimes as an announcement system for breaking changes, even when it's not really deprecating any functionality, so it didn't register with me based on that description, that a change of a default value would count. But I have no issue with it, good to know, ty!

felixfontein: For announcement-only deprecations (that have no practical effect on users since there is no code change) I think bugfix releases are totally fine, even though it might be strictly against semver :)

PR #299, opened 2022-08-28
add the changelog to the collection docsite

briantist:
##### SUMMARY

Since the collection's changelog is already RST, I had the idea of adding it to the docsite, so folks could more easily see the collection's changelog specifically.

Rendered:
- https://ansible-collections.github.io/community.hashi_vault/pr/299/collections/community/hashi_vault/index.html

---

I want to get feedback from the community and [DaWGs](https://github.com/ansible/community/issues/643) about:
- whether this is a good idea
- if the way I'm implementing it makes sense
- if we maybe want to provide a better/easier path for this via changes to the docs generation process
- if I've missed something entirely and the changelog is actually published on the docsite (though I would argue if so, it's been hard to find, and doing it this way will make it visible in local doc generation too)

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Once the new collection is published and the Zuul job is finished, add a release
 
 ## Release notes
 
-See the [changelog](https://github.com/ansible-collections/community.hashi_vault/tree/main/CHANGELOG.rst).
+See the [rendered changelog](https://ansible-collections.github.io/community.hashi_vault/branch/main/collections/community/hashi_vault/docsite/CHANGELOG.html) or the [raw generated changelog](https://github.com/ansible-collections/community.hashi_vault/tree/main/CHANGELOG.rst).
 
 ## FAQ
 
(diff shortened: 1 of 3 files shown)


mariolenz:
I think a changelog is an important part of the documentation and adding it to the docsite would make it more accessible. It's hard to find the changes to a collection you're interested in in the Ansible changelog. Since there is already the module (and some other) documentation on the docsite, why make people look in the GitHub repository for the changelog?

That said, I think it's a good idea. However, I don't know that much about docsite and how it works exactly. So there might be some downsides I am unable to see.

felixfontein (commented):
Looks good to me. It would probably be more cool if an existing changelog (in `changelog.yaml` format) is automatically included in the docsite by antsibull-docs. I might work on that soon...

briantist:
Thanks @felixfontein , that's the kind of first class support I was thinking could be nice, though probably opt-in since IIRC collections do have some options for handling changelogs differently, but yeah that'd be great! 

I will probably do it this way in the meantime, and switch if/when an integrated method is available.

felixfontein:
I created an issue in antsibull-docs for discussing this: https://github.com/ansible-community/antsibull-docs/issues/31

PR #303, opened 2022-09-17
`vault_token_create` -- use new hvac method

### LINKED ISSUE #301: Update `vault_token_create` orphan support to use new `hvac` method
##### SUMMARY

The method in `hvac` that can be used to create orphan tokens through the `/create-orphan` endpoint is deprecated for removal in v1, but had no replacement.

With https://github.com/hvac/hvac/pull/869 a replacement is being introduced for the 1.0.0 release.

We'll need to update `vault_token_create` to use this new method.

We will prefer this one, but also fall back to the old one, so that we keep compatibility with older versions of `hvac` for a bit.

The best-effort fallback to the method that requires root can be removed without deprecation; it was put in place as a potential fallback in case `hvac` 1.0.0 was released with no replacement, so with a replacement now there's no need to use it.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

vault_token_create

briantist:
##### SUMMARY

Fixes #301 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

vault_token_create

### DIFF
--- a/plugins/doc_fragments/token_create.py
+++ b/plugins/doc_fragments/token_create.py
@@ -15,7 +15,7 @@ options:
   orphan:
     description:
       - When C(true), uses the C(/create-orphan) API endpoint, which requires C(sudo) (but not C(root)) to create an orphan.
-      - With `hvac>=1.0.0`, requires collection version `>=3.3.0`.
+      - With C(hvac>=1.0.0), requires collection version C(>=3.3.0).
     type: bool
     default: false
   no_parent:
--- a/plugins/doc_fragments/token_create.py
+++ b/plugins/doc_fragments/token_create.py
@@ -15,15 +15,12 @@ options:
   orphan:
     description:
       - When C(true), uses the C(/create-orphan) API endpoint, which requires C(sudo) (but not C(root)) to create an orphan.
-      - Implies I(no_parent=true).
-      - B(NOTE:) as of this writing, the underlying endpoint in the C(hvac) library to support this is deprecated and scheduled for removal in C(v1.0.0).
-      - If I(orphan=true) and we cannot access the intended endpoint, the call will be attempted with the C(/create) endpoint, which requires root.
-      - If a replacement is provided in C(hvac), we will add support for it.
+      - With `hvac>=1.0.0`, requires collection version `>=3.3.0`.
     type: bool
     default: false
   no_parent:
     description:
-      - This option only has effect if used by a C(root) or C(sudo) caller, or in combination with I(orphan=true).
+      - This option only has effect if used by a C(root) or C(sudo) caller and I(orphan=false).
       - When C(true), the token created will not have a parent.
     type: bool
   no_default_policy:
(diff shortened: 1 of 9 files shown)


plugins/doc_fragments/token_create.py
```diff
@@ -15,15 +15,12 @@ class ModuleDocFragment(object):
...
-      - B(NOTE:) as of this writing, the underlying endpoint in the C(hvac) library to support this is deprecated and scheduled for removal in C(v1.0.0).
-      - If I(orphan=true) and we cannot access the intended endpoint, the call will be attempted with the C(/create) endpoint, which requires root.
-      - If a replacement is provided in C(hvac), we will add support for it.
+      - With C(hvac>=1.0.0), requires collection version C(>=3.3.0).
     type: bool
     default: false
   no_parent:
     description:
-      - This option only has effect if used by a C(root) or C(sudo) caller, or in combination with I(orphan=true).
+      - This option only has effect if used by a C(root) or C(sudo) caller and I(orphan=false).
```
acozine, line 23:
```suggestion
      - This option only has effect if used by a C(root) or C(sudo) caller and only when I(orphan=false).
```

acozine (commented):
The requirement wasn't clear to me on first reading. Is my suggestion is correct? If so, I think it clarifies what users must do.

plugins/modules/vault_token_create.py
```diff
@@ -192,17 +189,17 @@ def run_module():
...
         try:
-            response = client.create_token(orphan=True, **legacy_options)
+            # this method was added in hvac 1.0.0
+            # See: https://github.com/hvac/hvac/pull/869
+            response = client.auth.token.create_orphan(**orphan_options)
         except AttributeError:
-            module.warn("'create_token' method was not found. Attempting method that requires root privileges.")
+            # this method was removed in hvac 1.0.0
+            # See: https://github.com/hvac/hvac/issues/758
+            response = client.create_token(orphan=True, **orphan_options)
```
felixfontein, line 199: Shouldn't the above `try`/`except` be wrapped in another `try:`, so that the below `except:` is actually called if the above `client.create_token(...)` call raises an exception?

felixfontein (commented):
I cannot say much about the change itself, but I have a more generic comment :)

plugins/modules/vault_token_create.py
briantist: Yes, I believe you are correct, good catch, thank you!

briantist: https://github.com/ansible-collections/community.hashi_vault/pull/303/commits/64861315fe8a45249f441101f22ea6b0b896a8d3 adds the tests to catch this oversight (and this should be reflected in failed unit tests in CI). Next commit will have the fixed code making those tests pass.

felixfontein (approved):
As far as I can judge it, looks good. Obviously I haven't tested this (as I don't use Vault), but I guess you did :)

briantist:
@felixfontein @acozine  thank you for reviewing!

Issue #252, opened 2022-04-26
add tests for modules being included in the action group


briantist:
##### SUMMARY

I was thinking about how to prevent #251 in the future, and maybe trying to add tests for modules to ensure that they are in the action group. They would have to run only on core-2.12 and higher, was thinking I could set defaults for the group, then run the module in a way it should pass.

The catch is, that this would have to be added to new module's tests to be effective, so if you remember to add the test, you'll remember to add it to the action group (mission accomplished?) but if you forget to add the test, you'll also forget to add it to the action group, and the test will not catch it (mission failed).

I might have to look at this another way, maybe something a but more meta, like a general test that reads lists all the modules in the collection, parses `meta/runtime.yml`, and throws an error if any modules are not included. There'd be an exclude file somewhere so that modules that shouldn't be in the group can be excluded, that way the test will fail unless you either add the module to the action group, or add it to the exclusion list.

This kind of only works for as long as there is a single action group where most modules are expected to go. That will probably remain the case, but some revisiting will be needed if new action groups are created.

I suppose the same idea works: either add the module to _any_ group or mark it as excluded.

##### ISSUE TYPE
- Test Idea

##### ADDITIONAL INFORMATION

Related: #251

felixfontein:
> I might have to look at this another way, maybe something a but more meta, like a general test that reads lists all the modules in the collection, parses `meta/runtime.yml`, and throws an error if any modules are not included. There'd be an exclude file somewhere so that modules that shouldn't be in the group can be excluded, that way the test will fail unless you either add the module to the action group, or add it to the exclusion list.

:+1: I would do it like this.

PR #304, opened 2022-09-17
Add vault_kv2_delete module

### LINKED ISSUE #300: Add KV v2 delete functionality.
##### SUMMARY
I have a use case to be able to delete (versions) of kv2 data in Vault. I may be able to contribute to this feature.

Since the KV v2 API is a bit more complex than some of the other secret engines, I'm not sure if it would make sense to have `vault_kv2_{verb}` modules, or just generic `vault_{rest_api_verb}` modules (i.e. futureproofing for other features like list, metadata, destroy and undelete). 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
- vault_kv2_delete_version ?
- vault_delete ?

##### ADDITIONAL INFORMATION

- If `versions` not supplied, will default to latest version (same default as api)

```yaml
- name: Delete version 5 of a secret from kv2 with a different mount via the remote host
  community.hashi_vault.vault_kv2_delete_version:
    url: https://vault:8201
    path: path/to/secret
    versions: [5]

```

idwagner:
##### SUMMARY
Implementation of vault_kv2_delete module (hvac delete_latest_version_of_secret)

Fixes #300

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

vault_kv2_delete

##### ADDITIONAL INFORMATION

This implements the delete latest version functionality of a KV v2 secret in Vault. This is a WIP and is missing:
- Module docs
- Integration tests

### DIFF
--- a/plugins/modules/vault_kv2_delete.py
+++ b/plugins/modules/vault_kv2_delete.py
@@ -21,7 +21,6 @@ notes:
   - This module always reports C(changed) status because it cannot guarantee idempotence.
   - Use C(changed_when) to control that in cases where the operation is known to not change state.
 seealso:
-  - module: vault_kv2_delete
   - name: KV2 Secrets Engine
     description: Documentation for the Vault KV secrets engine, version 2.
     link: https://www.vaultproject.io/docs/secrets/kv/kv-v2
--- a/plugins/modules/vault_kv2_delete.py
+++ b/plugins/modules/vault_kv2_delete.py
@@ -1,9 +1,40 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2022, Isaac Wagner (@idwagner)
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
 
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 DOCUMENTATION = r'''
-module: vault_kv2_get
+module: vault_kv2_delete
+author:
+  - Isaac Wagner (@idwagner)
+short_description: Delete the latest version of a secret from HashiCorp Vault's KV version 2 secret store
+requirements:
+  - C(hvac) (L(Python library,https://hvac.readthedocs.io/en/stable/overview.html))
+  - For detailed requirements, see R(the collection requirements page,ansible_collections.community.hashi_vault.docsite.user_guide.requirements).
+description:
+  - Deletes the latest version of a secret from HashiCorp Vault's KV version 2 secret store.
+seealso:
+  - name: KV2 Secrets Engine
+    description: Documentation for the Vault KV secrets engine, version 2.
+    link: https://www.vaultproject.io/docs/secrets/kv/kv-v2
+extends_documentation_fragment:
+  - community.hashi_vault.connection
+  - community.hashi_vault.auth
+  - community.hashi_vault.engine_mount
+options:
+  engine_mount_point:
+    default: secret
+  path:
+    description:
+      - Vault KV path to be deleted.
+      - This is relative to the I(engine_mount_point), so the mount path should not be included.
+      - For kv2, do not include C(/data/) or C(/metadata/).
+    type: str
+    required: True
 '''
 
 
@@ -28,7 +59,7 @@ def run_module():
 
 
     argspec = HashiVaultModule.generate_argspec(
-        engine_mount_point=dict(type='str', default='kv'),
+        engine_mount_point=dict(type='str', default='secret'),
         path=dict(type='str', required=True)
     )
 
--- a/plugins/modules/vault_kv2_delete.py
+++ b/plugins/modules/vault_kv2_delete.py
@@ -21,7 +21,7 @@ notes:
   - This module always reports C(changed) status because it cannot guarantee idempotence.
   - Use C(changed_when) to control that in cases where the operation is known to not change state.
 seealso:
-  - module: vault_kv2_get
+  - module: vault_kv2_delete
   - name: KV2 Secrets Engine
     description: Documentation for the Vault KV secrets engine, version 2.
     link: https://www.vaultproject.io/docs/secrets/kv/kv-v2
@@ -130,7 +130,7 @@ def run_module():
         # Vault has two separate methods, one for delete latest version,
         # and delete specific versions.
         if module.check_mode:
-          response = {}
+            response = {}
         elif not versions:
             response = client.secrets.kv.v2.delete_latest_version_of_secret(
                 path=path, mount_point=engine_mount_point)
--- a/plugins/modules/vault_kv2_delete.py
+++ b/plugins/modules/vault_kv2_delete.py
@@ -38,10 +38,16 @@ options:
       - For kv2, do not include C(/data/) or C(/metadata/).
     type: str
     required: True
+  versions:
+    description:
+      - The version of the secret to delete.
+      - When omitted, the latest version of the secret is deleted.
+    type: list[int]
+    required: False
 '''
 
 EXAMPLES = """
-- name: Write a value to the cubbyhole via the remote host with userpass auth
+- name: Delete the latest version of the C(secret/mysecret) secret.
   community.hashi_vault.vault_kv2_delete:
     url: https://vault:8201
     path: secret/mysecret
@@ -50,14 +56,19 @@ EXAMPLES = """
     password: '{{ passwd }}'
   register: result
 
-- name: Display the result of the delete (this can be empty)
-  ansible.builtin.debug:
-    msg: "{{ result.data }}"
+- name: Delete versions 1 and 3 of the C(secret/mysecret) secret.
+  community.hashi_vault.vault_kv2_delete:
+    url: https://vault:8201
+    path: secret/mysecret
+    versions: [1, 3]
+    auth_method: userpass
+    username: user
+    password: '{{ passwd }}'
 """
 
 RETURN = """
 data:
-  description: The raw result of the write against the given path.
+  description: The raw result of the delete against the given path.
   returned: success
   type: dict
 """
@@ -84,7 +95,8 @@ def run_module():
 
     argspec = HashiVaultModule.generate_argspec(
         engine_mount_point=dict(type='str', default='secret'),
-        path=dict(type='str', required=True)
+        path=dict(type='str', required=True),
+        versions=dict(type='list', elements='int', required=False)
     )
 
     module = HashiVaultModule(
@@ -100,6 +112,7 @@ def run_module():
 
     engine_mount_point = module.params.get('engine_mount_point')
     path = module.params.get('path')
+    versions = module.params.get('versions')
 
     module.connection_options.process_connection_options()
     client_args = module.connection_options.get_hvac_connection_options()
@@ -111,8 +124,17 @@ def run_module():
     except (NotImplementedError, HashiVaultValueError) as e:
         module.fail_json(msg=to_native(e), exception=traceback.format_exc())
 
+
     try:
-        response = client.secrets.kv.v2.delete_latest_version_of_secret(path=path, mount_point=engine_mount_point)
+        # Vault has two separate methods, one for delete latest version,
+        # and delete specific versions.
+        if not versions:
+            response = client.secrets.kv.v2.delete_latest_version_of_secret(
+                path=path, mount_point=engine_mount_point)
+        else:
+            response = client.secrets.kv.v2.delete_secret_versions(
+                path=path, versions=versions, mount_point=engine_mount_point)
+
     except hvac.exceptions.Forbidden as e:
         module.fail_json(msg="Forbidden: Permission Denied to path ['%s']." % path, exception=traceback.format_exc())
 
--- a/plugins/modules/vault_kv2_delete.py
+++ b/plugins/modules/vault_kv2_delete.py
@@ -11,16 +11,17 @@ DOCUMENTATION = '''
 module: vault_kv2_delete
 author:
   - Isaac Wagner (@idwagner)
-short_description: Delete the latest version of a secret from HashiCorp Vault's KV version 2 secret store
+short_description: Delete one or more versions of a secret from HashiCorp Vault's KV version 2 secret store
 requirements:
   - C(hvac) (L(Python library,https://hvac.readthedocs.io/en/stable/overview.html))
   - For detailed requirements, see R(the collection requirements page,ansible_collections.community.hashi_vault.docsite.user_guide.requirements).
 description:
-  - Deletes the latest version of a secret from HashiCorp Vault's KV version 2 secret store.
+  - Delete one or more versions of a secret from HashiCorp Vault's KV version 2 secret store.
 notes:
   - This module always reports C(changed) status because it cannot guarantee idempotence.
   - Use C(changed_when) to control that in cases where the operation is known to not change state.
 seealso:
+  - module: vault_kv2_get
   - name: KV2 Secrets Engine
     description: Documentation for the Vault KV secrets engine, version 2.
     link: https://www.vaultproject.io/docs/secrets/kv/kv-v2
@@ -40,7 +41,7 @@ options:
     required: True
   versions:
     description:
-      - The version of the secret to delete.
+      - One or more versions of the secret to delete.
       - When omitted, the latest version of the secret is deleted.
     type: list
     elements: int
@@ -48,7 +49,7 @@ options:
 '''
 
 EXAMPLES = """
-- name: Delete the latest version of the C(secret/mysecret) secret.
+- name: Delete the latest version of the secret/mysecret secret.
   community.hashi_vault.vault_kv2_delete:
     url: https://vault:8201
     path: secret/mysecret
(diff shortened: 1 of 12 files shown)


plugins/modules/vault_kv2_delete.py
```diff
@@ -0,0 +1,71 @@
...
+
+    try:
+        module.authenticator.validate()
+        module.authenticator.authenticate(client)
+    except (NotImplementedError, HashiVaultValueError) as e:
+        module.fail_json(msg=to_native(e), exception=traceback.format_exc())
+
+    try:
+        raw = client.secrets.kv.v2.delete_latest_version_of_secret(path=path, mount_point=engine_mount_point)
+    except hvac.exceptions.Forbidden as e:
```
briantist, line 61: You may want to check for some more exceptions, depending on what `hvac` might raise/handle (best to check the source there to get an idea what to handle here).

tests/unit/plugins/modules/test_vault_kv2_delete.py
```diff
@@ -0,0 +1,154 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2022 Brian Scholer (@briantist)
```
briantist, line 2: I know it's a WIP but update this with your info.

And the method names will need to change from the kv2_get names as well.

briantist (commented):
Thank you @idwagner ! This is a great start :)

tests/unit/plugins/modules/test_vault_kv2_delete.py
idwagner: done :)

plugins/modules/vault_kv2_delete.py
idwagner: Looks like hvac does not do any error handling for this. I also peeked at the Vault source, and it doesn't do anything special in the delete itself, just passing through any errors it gets such as any errors updating the metadata. If theres anything else you can think of, I'd be happy to add it.

plugins/modules/vault_kv2_delete.py
```diff
@@ -0,0 +1,71 @@
...
+        module.fail_json(msg="Forbidden: Permission Denied to path ['%s']." % path, exception=traceback.format_exc())
+
+    module.exit_json(changed=True, response_code=raw.status_code)
```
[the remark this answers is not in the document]
idwagner: I've added the vault_write logic along with the associated pertinent docs on the return.

plugins/modules/vault_kv2_delete.py
```diff
@@ -0,0 +1,71 @@
...
+    HVAC_IMPORT_ERROR = traceback.format_exc()
+else:
+    HVAC_IMPORT_ERROR = None
+    HAS_HVAC = True
+
+def run_module():
+
+
+    argspec = HashiVaultModule.generate_argspec(
+        engine_mount_point=dict(type='str', default='kv'),
```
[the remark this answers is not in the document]
idwagner: done :)

idwagner:
@briantist few questions,

Is there any convention for adding policies to the integration tests by path or key? Should I create a policy on a new kv path for these tests rather than modifying an existing path permission?

Also, I'm thinking maybe I should rename `vault_kv2_delete` to `vault_kv2_delete_latest_version` to more closely match hvac's methods. Further development could have modules like `vault_kv2_delete_version` `vault_kv2_undelete` and `vault_kv2_destroy` etc...

briantist:
> @briantist few questions,
> 
> Is there any convention for adding policies to the integration tests by path or key? 

Yes, have a look at [this file for the tasks](https://github.com/ansible-collections/community.hashi_vault/blob/main/tests/integration/targets/setup_vault_configure/tasks/configure.yml), and [this file for the policy definitions](https://github.com/ansible-collections/community.hashi_vault/blob/main/tests/integration/targets/setup_vault_configure/vars/main.yml).

> Should I create a policy on a new kv path for these tests rather than modifying an existing path permission?

Feel it out and see what makes sense. A new path is probably the way to go, where the "setup" parts of the integration tests would create the secret(s) to be deleted in the tests.

Also have a look at the [test plugins](https://github.com/ansible-collections/community.hashi_vault/tree/main/tests/integration/targets/setup_vault_test_plugins), in case a new one needs to be added there for any reason (for example for creating/deleting the secrets).

> Also, I'm thinking maybe I should rename `vault_kv2_delete` to `vault_kv2_delete_latest_version` to more closely match hvac's methods. Further development could have modules like `vault_kv2_delete_version` `vault_kv2_undelete` and `vault_kv2_destroy` etc...

I think I'm still leaning toward a single `vaukt_kv2_delete` that will have the ability to optionally delete a specific version or versions depending on how the options are set. There can still be other modules like `vault_kv2_undelete` and `vault_kv2_destroy`. Any reason in particular you're thinking of going back the other way?

tests/unit/plugins/modules/test_vault_kv2_delete.py
briantist: Method names look good, but it's still my name in the copyright ;)

idwagner:
@briantist I believe this is about ready with the integration tests in place. Two of the tests failed with errors that lead me to believe they are re-using a vault session from a previous test? I setup some seed data in `setup_vault_configure` assuming that would be run during each integration test, however the failed tests don't match that initial seed data.

https://github.com/idwagner/community.hashi_vault/blob/a0f8e7b4597408e8ad330a612e4a783d9b3db6a1/tests/integration/targets/setup_vault_configure/tasks/configure.yml#L100-L107

briantist:
> @briantist I believe this is about ready with the integration tests in place. Two of the tests failed with errors that lead me to believe they are re-using a vault session from a previous test? I setup some seed data in `setup_vault_configure` assuming that would be run during each integration test, however the failed tests don't match that initial seed data.
> 
> https://github.com/idwagner/community.hashi_vault/blob/a0f8e7b4597408e8ad330a612e4a783d9b3db6a1/tests/integration/targets/setup_vault_configure/tasks/configure.yml#L100-L107

Ah! I'm sorry you hit that error, but I am happy to see that those tests are failing in exactly the right situation they were designed to!

> Two of the tests failed with errors that lead me to believe they are re-using a vault session from a previous test?

They are doing that, intentionally. The key is that I want the tests to be able to run against a local instance of Vault that doesn't need setup on every single run, because it speeds up local iteration.

So in your case, the problem, is that after you create those secret versions and delete some of them, those versions do not exist again. And if you tried to insert new versions, they would not "fill in" those missing ones either they would continue adding new versions to the end (versions 6, 7, 8, etc.).

For a test like this, instead of setting up that secret in the "test-wide" re-usable secrets, you should set it up in `tests/integration/targets/module_vault_kv2_delete/tasks/module_vault_kv2_delete_setup.yml` in the configuration tasks. In here, you should:
1. Delete the existing secret completely if it exists (this must succeed whether it exists or not)
2. Create the secret and all of its versions.

It should be removed from the general vault configure target.

galaxy.yml
```diff
@@ -7,6 +7,7 @@ readme: README.md
 authors:
   - Julie Davila (@juliedavila) <julie(at)davila.io>
   - Brian Scholer (@briantist)
+  - Isaac Wagner (@idwagner)
```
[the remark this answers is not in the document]
idwagner: Nor worries, I will remove that 👍

idwagner:
It looks like the change corrected one of the tests, but other timed out. Can that test be re-run without a push?

plugins/modules/vault_kv2_delete.py
```diff
@@ -0,0 +1,165 @@
...
+
+    argspec = HashiVaultModule.generate_argspec(
+        engine_mount_point=dict(type='str', default='secret'),
+        path=dict(type='str', required=True),
+        versions=dict(type='list', elements='int', required=False)
+    )
+
+    module = HashiVaultModule(
+        argument_spec=argspec,
+        supports_check_mode=True
```
briantist, line 105: At the moment, the module does not account for check mode, and so with this set to `True`, check mode would delete secrets 😱 

One issue currently (same reason status is always `changed`) is that without checking the current status of the secret, check mode has nothing to do but return `changed`.

I'm mulling over the idea of using metadata reads to get current versions status/secret existence to be able to solve both of those, but their use should be optional (a client may have permission to delete but not to hit those other endpoints).

I'll look at this again tomorrow and think about whether that makes sense at all, and if so whether it should be in this PR or a follow-up.

The first thing we should do is add an integration test and unit test ([example](https://github.com/ansible-collections/community.hashi_vault/blob/main/tests/unit/plugins/modules/test_vault_login.py#L98)) for check mode before making any additional changes, and those tests should fail right now.

tests/integration/targets/module_vault_kv2_delete/tasks/module_vault_kv2_delete_setup.yml
```diff
@@ -0,0 +1,29 @@
...
+        mount_point: '{{ vault_kv2_mount_point }}'
+        path: '{{ vault_kv2_versioned_path }}/secret6'
+
+    - name: Set up a multi versioned secret for delete (v2)
+      vault_ci_kv_put:
+        version: 2
+        mount_point: '{{ vault_kv2_mount_point }}'
+        path: '{{ vault_kv2_versioned_path }}/secret6'
+        secret:
+          v: value{{item}}
```
briantist, line 28:
```suggestion
          v: value{{ item }}
```

briantist (commented):
See the comment about check mode.

The rest are mostly small changes.

In case you haven't seen it before, you can commit multiple suggestions from review at once, if you view the comments in the Files tab. Then, there will be an option to add to batch, and then you can commit the batch at once.

plugins/modules/vault_kv2_delete.py
```diff
@@ -0,0 +1,165 @@
...
+short_description: Delete the latest version of a secret from HashiCorp Vault's KV version 2 secret store
+requirements:
+  - C(hvac) (L(Python library,https://hvac.readthedocs.io/en/stable/overview.html))
+  - For detailed requirements, see R(the collection requirements page,ansible_collections.community.hashi_vault.docsite.user_guide.requirements).
+description:
+  - Deletes the latest version of a secret from HashiCorp Vault's KV version 2 secret store.
+notes:
+  - This module always reports C(changed) status because it cannot guarantee idempotence.
+  - Use C(changed_when) to control that in cases where the operation is known to not change state.
+seealso:
```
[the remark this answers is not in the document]
briantist: My original suggestion was to link to `vault_kv2_get`, then you changed it to `vault_kv2_delete` in https://github.com/ansible-collections/community.hashi_vault/pull/304/commits/9cafc720ed62d9dd923b29460b9743d21c33d292 , then deleted it in https://github.com/ansible-collections/community.hashi_vault/pull/304/commits/1b3aeb181a03610759d0d2ea738034fbe100eb6c as being self-referential 😅

I still think a link to `vault_kv2_get` is a good idea though. I find it's nice to link to other kv2 modules, do you prefer not having it?

```suggestion
seealso:
  - module: vault_kv2_get
```

plugins/modules/vault_kv2_delete.py
```diff
@@ -0,0 +1,167 @@
...
+  - Isaac Wagner (@idwagner)
+short_description: Delete one or more versions of a secret from HashiCorp Vault's KV version 2 secret store
+requirements:
+  - C(hvac) (L(Python library,https://hvac.readthedocs.io/en/stable/overview.html))
+  - For detailed requirements, see R(the collection requirements page,ansible_collections.community.hashi_vault.docsite.user_guide.requirements).
+description:
+  - Delete one or more versions of a secret from HashiCorp Vault's KV version 2 secret store.
+notes:
+  - This module always reports C(changed) status because it cannot guarantee idempotence.
+  - Use C(changed_when) to control that in cases where the operation is known to not change state.
```
briantist, line 22:
```suggestion
  - Use C(changed_when) to control that in cases where the operation is known to not change state.
  - In check mode, the module returns C(changed) status without contacting Vault. Consider using M(community.hashi_vault.vault_kv2_get) to verify the existence of the secret first.
```

plugins/modules/vault_kv2_delete.py
```diff
@@ -0,0 +1,167 @@
...
+    path: secret/mysecret
+    versions: [1, 3]
+    auth_method: userpass
+    username: user
+    password: '{{ passwd }}'
+"""
+
+RETURN = """
+data:
+  description: The raw result of the delete against the given path.
```
briantist, line 72:
```suggestion
  description: 
    - The raw result of the delete against the given path.
    - This is usually empty, but may contain warnings or other information.
```

plugins/modules/vault_kv2_delete.py
briantist: Thank you, I think for now the empty result with no Vault connection is fine.

The unit tests are great! Please also add an integration test for check mode; it should be easy, just checking that the result `is changed` and that the deletion did not happen.

Even though the unit test covers this a little more thoroughly, it's always good to have it covered by integration too just to have the code actually invoked by ansible.

tests/integration/targets/module_vault_kv2_delete/tasks/module_vault_kv2_delete_setup.yml
```diff
@@ -0,0 +1,29 @@
...
+      vault_ci_token_create:
+        policies: [test-policy]
+      register: user_token_cmd
+
+- name: Configuration tasks
+  module_defaults:
+    vault_ci_kv2_destroy_all: '{{ vault_plugins_module_defaults_common }}'
+    vault_ci_kv_put: '{{ vault_plugins_module_defaults_common }}'
+  block:
+
```
briantist, line 16: nit
```suggestion
```

tests/integration/targets/module_vault_kv2_delete/tasks/module_vault_kv2_delete_test.yml
```diff
@@ -0,0 +1,175 @@
+---
+
```
briantist, line 2: nit
```suggestion
```

plugins/modules/vault_kv2_delete.py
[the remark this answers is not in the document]
idwagner: I don't know what happened there 😃 I agree `vault_kv2_get` should be referenced.

tests/integration/targets/module_vault_kv2_delete/tasks/module_vault_kv2_delete_test.yml
```diff
@@ -0,0 +1,220 @@
...
+              - "kv2_result['result']['data']['versions']['4']['deletion_time'] is search(regex_secret_version_not_deleted)"
+              - "kv2_result['result']['data']['versions']['5']['deletion_time'] is search(regex_secret_version_not_deleted)"
+            fail_msg: 'Test Seed value did not contain expected data.'
+
+        - name: Try kv2 delete latest version in check mode
+          register: kv2_result
+          community.hashi_vault.vault_kv2_delete:
+            path: "{{ vault_kv2_versioned_path }}/secret6"
+          check_mode: true
+
```
briantist, line 60: let's add assertions for the output result of the deletions too; the intention will be to test these after each call to the delete module.

```suggestion

        - assert:
            that: &delete_assertions
              - kv2_result is changed
              - kv2_result.data == {}

```

tests/integration/targets/module_vault_kv2_delete/tasks/module_vault_kv2_delete_test.yml
```diff
@@ -0,0 +1,220 @@
...
+            fail_msg: 'Secret version was deleted while in check mode.'
+
+        - name: Try kv2 delete version 1 and 3
+          register: kv2_result
+          community.hashi_vault.vault_kv2_delete:
+            path: "{{ vault_kv2_versioned_path }}/secret6"
+            versions:
+              - 1
+              - 3
+        - assert:
```
briantist, line 108: spacing is a nit

But something is strange here. I don't understand how this assertion is passing. `kv2_result` should contain nothing but `changed=true` and `data={}`, so how are the basics assertions not failing?

```suggestion

        - assert:
```

briantist: Ah! I have figured this out. We cannot use YAML anchors like this.

With the following:
```yaml
- assert:
    that: &a
      - 1
      - 2
```

The anchor is referring to the entire list value: `[1, 2]`.

When using it like this:
```yaml
- assert:
    that:
      - *a
      - b
```

Then, what we have for the list of assertions then is actually this: `[[1,2], "b"]`

So the first list item is a non-empty list which always evaluates to true.

For dictionaries, YAML anchors have the merge operator `<<` but there is no such thing for list.

We don't have a lot of good options for this. Either duplicate all of the "basic" assertions, or use vars in a way like this:

```
- name: some block
  vars:
    basic_assertions: &basics
      - 1
      - 2
 
- name: simple case - basics only - option 1
  assert:
    that: *basics

- name: simple case - basics only - option 2
  assert:
    that: '{{ basic_assertions }}'

- name: complex 1 - literals
  assert:
    that: "{{ basic_assertions + ['condition 1', 'condition 2'] }}"

- name: complex 2 - vars
  vars:
    these_assertions:
      - condition 1
      - condition 2
  assert:
    that: '{{ basic_assertions + these_assertions }}'
```

briantist:
There is one more thing I forgot (#252 🤦): please add this module to the action group:
- https://github.com/ansible-collections/community.hashi_vault/blob/main/meta/runtime.yml#L4

tests/integration/targets/module_vault_kv2_delete/tasks/module_vault_kv2_delete_test.yml
idwagner: I feel its a lot more readable to just duplicate the lines from the anchor where its needed (which is only a handful).. Cleaning up some of that now and will push something up once i run some local tests

briantist (approved):

briantist:
@idwagner thanks so much for this contribution! It's now released in [version 3.4.0](https://github.com/ansible-collections/community.hashi_vault/releases/tag/3.4.0)! 🥳

Issue #260, opened 2022-04-30
Add filter docs


briantist:
##### SUMMARY
Sidecar docs and docs for filter/test plugins are now available:
- https://github.com/ansible-collections/community.dns/pull/99
- https://github.com/ansible-collections/community.routeros/pull/87
- https://github.com/ansible-collections/community.sops/pull/99
- https://github.com/ansible-collections/community.general/pull/4597

Let's gooooooo!!

##### ISSUE TYPE
- Documentation Report

briantist:
Originally thought about deprecating the filter guide but may leave it for now since it can go into greater detail and I can't easily fit all that detail into a standard documentation page.

PR #306, opened 2022-09-25
Docs/build update stuff

briantist:
##### SUMMARY

Some stuff taken from #260 

- update the workflow dependency
- update the preview stuff (committed `antsibull-docs sphinx-init` output)
- exclude sphinx `.post0` release due to https://github.com/ansible-community/antsibull-docs/issues/39

Merging will fix the currently broken docs build; it will fail in this PR in the "base" build due this not being in `main` yet.
Validate ~should pass~ will not pass because with the change to use `2.14` for docs builds, the missing filter documentation is now a fatal error.

~I'm going to merge these changes first so that the base docs build in #260 can pass.~

Actually.. the base build will still fail because of the missing filter docs I suppose.. oof. Am going to just close this and do everything in the other one I guess.

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/.github/workflows/docs-push.yml
+++ b/.github/workflows/docs-push.yml
@@ -38,7 +38,7 @@ jobs:
     if: github.repository == 'ansible-collections/community.hashi_vault'
     permissions:
       contents: read
-    needs: [build-docs]
+    needs: [validate-docs, build-docs]
     name: Publish Ansible Docs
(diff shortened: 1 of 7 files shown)


briantist:
hm maybe I was wrong about the build, let's merge this and see

PR #317, opened 2022-11-01
update `token_validate` default value from `true` to `false`

### LINKED ISSUE #248: The default value of `token_validate` will change from `True` to `False` in version `4.0.0`

briantist:
##### SUMMARY

Resolves #248

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/changelogs/fragments/248-token_validate-default.yml
+++ b/changelogs/fragments/248-token_validate-default.yml
@@ -1,0 +1,3 @@
+---
+breaking_changes:
+  - auth - the default value for ``token_validate`` has changed from ``true`` to ``false``, as previously announced (https://github.com/ansible-collections/community.hashi_vault/issues/248).
--- a/plugins/doc_fragments/auth.py
+++ b/plugins/doc_fragments/auth.py
@@ -54,9 +54,8 @@ class ModuleDocFragment(object):
         description:
           - For token auth, will perform a C(lookup-self) operation to determine the token's validity before using it.
           - Disable if your token does not have the C(lookup-self) capability.
-          - The default value is C(true).
-          - The default value will change to C(false) in version 4.0.0.
         type: bool
+        default: false
         version_added: 0.2.0
       username:
         description: Authentication user name.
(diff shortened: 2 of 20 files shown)


felixfontein (approved):

briantist:
@felixfontein thanks for reviewing this too!

PR #322, opened 2022-11-04
`vault_kv2_get` lookup - change `engine_mount_point` default to `secret`

### LINKED ISSUE #279: The default for `vault_kv2_get` lookup's `engine_mount_point` will change to `secret` in version `4.0.0`

briantist:
##### SUMMARY

Fixes #279 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

vault_kv2_get lookup

### DIFF
--- a/changelogs/fragments/279-vault_kv2_get-lookup-mount-default.yml
+++ b/changelogs/fragments/279-vault_kv2_get-lookup-mount-default.yml
@@ -1,0 +1,3 @@
+---
+breaking_changes:
+  - vault_kv2_get lookup - as previously announced, the default value for ``engine_mount_point`` in the ``vault_kv2_get`` lookup has changed from ``kv`` to ``secret`` (https://github.com/ansible-collections/community.hashi_vault/issues/279).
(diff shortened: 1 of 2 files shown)


briantist:
The coverage "failure" is unavoidable without adding unrelated coverage. The file being changed  has 100% coverage, so removing some lines, while leaving the file with 100% coverage, is still ever so slightly reducing the coverage of the project as a whole (by 0.01%).

felixfontein (approved):

briantist:
@felixfontein thanks for reviewing this!

PR #333, opened 2022-11-19
[DNM] try improved container mgmt with macos

briantist:
##### SUMMARY

TESTING

Try out new container stuff to see if works with macos again (since 2.13 ansible-test didn't work with macos in CI anymore and we've been pinning it to 2.12).
- https://github.com/ansible/ansible/pull/78550

##### ISSUE TYPE

- CI Pull Request

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -424,7 +424,6 @@ jobs:
           retention-days: 1
 
   upload-coverage:
-    if: false
     needs:
       - sanity
       - units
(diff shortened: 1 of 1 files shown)


.github/workflows/ansible-test.yml
```diff
@@ -364,6 +368,13 @@ jobs:
         # sometimes this hangs forever waiting for an IP
         uses: ./.github/actions/macos-docker
 
+      - name: cgroups, see groups run, run groups run (please run!)
+        if: ${{ startsWith(matrix.runner, 'macos') }}
+        run: |
```
mattclay, line 373: This appears to be running on the macOS host instead of the Docker host.

briantist: yeah... I'm not sure how I would run that in the docker VM 🤔

PR #335, opened 2022-12-03
chores: use ubuntu 20.04 to unbreak ci

lowlydba:
https://github.com/ansible-collections/news-for-maintainers/discussions/29

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -33,7 +33,7 @@ jobs:
     strategy:
       matrix:
         runner:
-          - ubuntu-latest
+          - ubuntu-20.04
         test_container:
           - default
         ansible:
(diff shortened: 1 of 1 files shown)


briantist:
Why was the exclude change reverted in [71857ff](https://github.com/ansible-collections/community.hashi_vault/pull/335/commits/71857ff53d88a8980506fd6b159f772e9b8d275d) ? We do actually want that, could you restore?

lowlydba:
> Why was the exclude change reverted in [71857ff](https://github.com/ansible-collections/community.hashi_vault/pull/335/commits/71857ff53d88a8980506fd6b159f772e9b8d275d) ? We do actually want that, could you restore?

I don't remember 😨

briantist:
With the ansible-test updates backported to 2.14 & 2.13, I think this should now only explicitly use 20.04 in <=2.12 ; but it also seems like I have other problems in CI that I need to address. I may make a separate PR to fix it all at once when I'm back home and able to dedicate time to it. Let's keep this PR open for now just in case, thanks for putting this in anyway!

PR #342, opened 2023-01-08
Docs/fix attributes

### LINKED ISSUE #325: Update `check_mode` attributes to correct `description` and add `details`
##### SUMMARY

When I added attributes in #320 I didn't quite use `check_mode` correctly.
The `description` should be a general description of the attribute (what to expect from check mode), and details about how the particular module implements it should be added in a list under `details`, which will render underneath the `support` field in the docs.

Should update these to correct them.

Example: 
- https://github.com/ansible-collections/community.hrobot/pull/65

##### ISSUE TYPE
- Documentation Report

briantist:
##### SUMMARY

Fixes #325 

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/plugins/modules/vault_login.py
+++ b/plugins/modules/vault_login.py
@@ -24,6 +24,7 @@ DOCUMENTATION = """
     - ref: community.hashi_vault.vault_login_token filter <ansible_collections.community.hashi_vault.vault_login_token_filter>
       description: The official documentation for the C(community.hashi_vault.vault_login_token) filter plugin.
   extends_documentation_fragment:
+    - community.hashi_vault.attributes
     - community.hashi_vault.attributes.action_group
     - community.hashi_vault.connection
     - community.hashi_vault.auth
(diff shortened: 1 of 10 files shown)


plugins/modules/vault_login.py
```diff
@@ -40,10 +41,10 @@
   attributes:
     check_mode:
       support: partial
-      description:
-        - "In check mode, this module will not perform a login, and will instead return a basic structure with an empty token.
+      details:
+        - In check mode, this module will not perform a login, and will instead return a basic structure with an empty token.
           However this may not be useful if the token is required for follow on tasks.
-          It may be better to use this module with C(check_mode=no) in order to have a valid token that can be used."
+        - It may be better to use this module with C(check_mode=no) in order to have a valid token that can be used.
```
felixfontein, line 47: I'm not sure whether you already updated the documentation to use `true`/`false` in other places (or whether you want to do that), but here is a place where `yes`/`no` is still in use:
```suggestion
        - It may be better to use this module with C(check_mode=false) in order to have a valid token that can be used.
```

plugins/modules/vault_token_create.py
```diff
@@ -39,10 +40,10 @@
   attributes:
     check_mode:
       support: partial
-      description:
+      details:
         - In check mode, this module will not create a token, and will instead return a basic structure with an empty token.
           However, this may not be useful if the token is required for follow on tasks.
-          It may be better to use this module with I(check_mode=no) in order to have a valid token that can be used.
+        - It may be better to use this module with I(check_mode=no) in order to have a valid token that can be used.
```
felixfontein, line 46: Same here.
```suggestion
        - It may be better to use this module with I(check_mode=false) in order to have a valid token that can be used.
```

felixfontein (approved):
Besides this, LGTM.

felixfontein (approved):

PR #343, opened 2023-01-13
Add plugins to list objects in Vault

### LINKED ISSUE #295: Feature Request: Add possibility to get list of secrets from path
##### SUMMARY

We add users to VMs and the option to pick up all secrets from some path would be good.

I've checked `hvac` python module and found that it can do this:
```python
list_response = client.secrets.kv.v2.list_secrets(
    path='hvac',
)
```
I think this option can help and other cases. If I can do this with the actual version of `hashi_vault`, could you please share the information about this?
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

hashi_vault lookup
##### ADDITIONAL INFORMATION

```yaml

```

tomkivlin:
##### SUMMARY

Two plugins: one lookup, one module to list objects from a given path.
e.g.

* list all secrets in a given path
* list all access policies
* list all userpass users

Fixes #295 

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

vault_list (lookup and module)


briantist (commented):
@tomkivlin this a fantastic first contribution, thank you! 

There's some inline comments that are mostly small things.

The vault action group in `meta/runtime.yml` also needs the module added there:
https://github.com/ansible-collections/community.hashi_vault/blob/main/meta/runtime.yml#L4

The only other thing that I think would be nice is a couple of extra examples for using the lookup in a way where you loop over the actual thing being listed. Since this is the "generic" list lookup, you are (correctly!) returning the raw result, which is always a dictionary, but somewhere inside the dictionary (possibly nested) is a list.

So for example, an example showing how to loop over a list of policies might show a loop like this:

```yaml
loop: "{{ query('community.hashi_vault.vault_list', 'sys/policies/acl').keys }}"
```

and an example for kv2 might look like this:

```yaml
loop: "{{ query('community.hashi_vault.vault_list', 'secret/metadata/path').data.keys }}"
```

These types of examples that show how to do specific things tend to be highly used, and in fact sometimes people try to do these operations, get failures, and open issues about it, even if they only needed to tweak their path, or massage the output data.

---

One last thing just in case you haven't seen it before. Several of my comments have suggested edits, which you can commit from the comment. If you click that on each comment separately, they will all be separate tiny commits that make the CI take forever.

If you view the comments on the "Files" tab instead, you'll see an option to "Add suggestion to batch" and then you click wait until you've added all the suggestions you want to keep, and commit them at once.

You can also make any changes including changes to things I suggested, locally on your branch instead.

Thanks again! Really looking forward to adding these to the collection.

plugins/modules/vault_list.py
```diff
@@ -0,0 +1,135 @@
...
+
+- name: List access policies from Vault via the remote host
+  community.hashi_vault.vault_list:
+    url: https://vault:8201
+    path: sys/policies/acl
+  register: policies
+
+- name: Display the role IDs
+  ansible.builtin.debug:
+    msg: "{{ policies.data }}"
```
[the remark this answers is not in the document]
tomkivlin: Yes, you're right - thanks!

tests/integration/targets/lookup_vault_list/tasks/lookup_vault_list_test.yml
```diff
@@ -0,0 +1,98 @@
...
+          - '{{ vault_kv2_api_list_path }}'
+          - '{{ vault_policy_api_list_path }}'
+        list_results: "{{ lookup('community.hashi_vault.vault_list', *paths, **kwargs) }}"
+      assert:
+        that:
+          - list_results | type_debug == 'list'
+          - item | type_debug == 'dict'
+          - "'data' in item"
+          - "'keys' in item['data']"
+          # - "'value' in item['data']['keys']"
```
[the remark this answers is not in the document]
tomkivlin: yeah good idea, added

tomkivlin:
> So for example, an example showing how to loop over a list of policies might show a loop like this:
> 
> ```yaml
> loop: "{{ query('community.hashi_vault.vault_list', 'sys/policies/acl').keys }}"
> ```
> 
> and an example for kv2 might look like this:
> 
> ```yaml
> loop: "{{ query('community.hashi_vault.vault_list', 'secret/metadata/path').data.keys }}"
> ```
> 
> (**EDIT** I just realized fixing the above path that the examples should also have their "secret" listing paths updates to include `metadata` and some path)
> 
> These types of examples that show how to do specific things tend to be highly used, and in fact sometimes people try to do these operations, get failures, and open issues about it, even if they only needed to tweak their path, or massage the output data.

Thanks - great idea! However, I have come across some unexpected (to me) behaviour, possibly because 'keys' is a built-in method - hope my alternatives make sense. (In https://github.com/tomkivlin/community.hashi_vault/commit/cd583f9b8c8985a792deac010c295aa2a8cbb33a).

tests/unit/plugins/modules/test_vault_list.py
```diff
@@ -0,0 +1,157 @@
...
+def _combined_options(**kwargs):
+    opt = _connection_options()
+    opt.update(_sample_options())
+    opt.update(kwargs)
+    return opt
+
+
+@pytest.fixture
+def kv1_get_response(fixture_loader):
+    return fixture_loader('kv1_get_response.json')
```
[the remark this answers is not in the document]
tomkivlin: Updated in https://github.com/tomkivlin/community.hashi_vault/commit/148a3b1d6b60cb08be14c26d7c348df356b7993f

tests/unit/plugins/lookup/test_vault_list.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@pytest.fixture
+def vault_list_lookup():
+    return lookup_loader.get('community.hashi_vault.vault_list')
+
+
+@pytest.fixture
+def kv1_get_response(fixture_loader):
+    return fixture_loader('kv1_get_response.json')
```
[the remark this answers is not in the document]
tomkivlin: Updated in https://github.com/tomkivlin/community.hashi_vault/commit/148a3b1d6b60cb08be14c26d7c348df356b7993f

briantist:
> Thanks - great idea! However, I have come across some unexpected (to me) behaviour, possibly because 'keys' is a built-in method - hope my alternatives make sense. (In [tomkivlin@cd583f9](https://github.com/tomkivlin/community.hashi_vault/commit/cd583f9b8c8985a792deac010c295aa2a8cbb33a)).

ah! yes that makes sense, I might tweak it further with jinja in a way that will work even with multiple terms, I'll give it a closer look

> @briantist thanks for your comprehensive feedback on this, really helpful. I've added some updates, hopefully that covers it all?

Pretty much yeah, I see a little bit I'd like to update in the fixture tests, and we're missing a tiny bit of coverage that I want to fill. I aim for 100% coverage on everything new, and we're missing just one line I think (a 404/missing path response).

I have some time today so I'm going to pull this down and see if I button up these last little things, and then give it another over. I'll push something up soon.

briantist:
hey @tomkivlin , it seems I cannot push my changes to your fork. There should be a checkbox on this PR in the right column that allows you to grant maintainers the ability to push changes ([see the last screenshot here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)).

Would you mind ensuring that is checked?

If you prefer not to enable that, that's ok too, I can send a `.patch` or `.diff` that you can apply locally to push. Let me know!

briantist:
recheck

briantist:
@tomkivlin would you take a look at the commits I've pushed up and see if everything looks ok to you? Let me know if there are any questions or things I missed.

There's some weirdness zuul right now (ansible third party check), but that's because of anything in this PR as far as I can tell.

recheck

briantist:
recheck

briantist:
recheck

plugins/modules/vault_list.py
```diff
@@ -38,7 +38,7 @@
 - name: List kv2 secrets from Vault via the remote host with userpass auth
   community.hashi_vault.vault_list:
     url: https://vault:8201
-    path: secret
+    path: secret/metadata/path
```
tomkivlin, line 41: I got it wrong to start with - the path for kv2 needs to be `mount_point/metadata` so `secret/metadata` in this example. 
Both `secret` (i.e. the mount_point) and `secret/metadata/secret_name` fail for the list operation.

I will update in following commit.

plugins/lookup/vault_list.py
```diff
@@ -42,15 +42,15 @@
 - name: Perform multiple list operations with a single Vault login
   vars:
     paths:
-      - secret
+      - secret/metadata/path
```
tomkivlin, line 45: Needs to be 'secret/metadata' - will update.

plugins/lookup/vault_list.py
```diff
@@ -42,15 +42,15 @@
...
       - sys/policies/acl
   ansible.builtin.debug:
     msg: "{{ lookup('community.hashi_vault.vault_list', *paths, auth_method='userpass', username=user, password=pwd) }}"
 
 - name: Perform multiple list operations with a single Vault login in a loop
   vars:
     paths:
-      - secret
+      - secret/metadata/path
```
tomkivlin, line 53: Needs to be 'secret/metadata' - will update.

plugins/lookup/vault_list.py
```diff
@@ -64,22 +64,58 @@
   ansible.builtin.debug:
     msg: '{{ item }}'
   with_community.hashi_vault.vault_list:
-    - secret
+    - secret/metadata/path
```
tomkivlin, line 67: Needs to be 'secret/metadata' - will update.

tests/integration/targets/lookup_vault_list/tasks/lookup_vault_list_test.yml
```diff
@@ -75,7 +75,7 @@
     - assert:
         that:
           - test_inexistent is failed
-          - test_inexistent.msg is search("Permission Denied")  # This is the error received when listing a non-existent path
+          - test_inexistent.msg is search("doesn't seem to exist")
```
tomkivlin, line 78: I think we need to reverse this change - this is an example error I get for a list operation on an inexistent path:

```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Forbidden: Permission Denied to path 'vpn-users_inexistent/metadata'."}
```

plugins/lookup/vault_list.py
tomkivlin: Done in https://github.com/ansible-collections/community.hashi_vault/pull/343/commits/857568ab7e72c9826cf56b5bb3691a4d950433d3

plugins/lookup/vault_list.py
tomkivlin: Done in https://github.com/ansible-collections/community.hashi_vault/pull/343/commits/857568ab7e72c9826cf56b5bb3691a4d950433d3

plugins/lookup/vault_list.py
```diff
@@ -33,7 +33,7 @@
 EXAMPLES = """
 - name: List all secrets at a path
   ansible.builtin.debug:
-    msg: "{{ lookup('community.hashi_vault.vault_list', 'secret', url='https://vault:8201') }}"
+    msg: "{{ lookup('community.hashi_vault.vault_list', 'secret/metadata/path', url='https://vault:8201') }}"
```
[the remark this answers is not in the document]
tomkivlin: Done in https://github.com/ansible-collections/community.hashi_vault/pull/343/commits/857568ab7e72c9826cf56b5bb3691a4d950433d3

plugins/modules/vault_list.py
tomkivlin: Done in https://github.com/ansible-collections/community.hashi_vault/pull/343/commits/857568ab7e72c9826cf56b5bb3691a4d950433d3

tests/integration/targets/lookup_vault_list/tasks/lookup_vault_list_test.yml
briantist: If you don't have permission to a path in Vault, it will generally give you a 403 instead of a 404, to avoid leaking information about whether something exists at a given path.

But if you have permission (via policy), it gives the 404 correctly, so we do still need to account for that change. The code already did, but that code was never reached in the tests.

The reason the tests also showed 403 is that the policy didn't include permissions for the path being checked. That's one of the things I updated along with this.

Though one thing I noticed is that if the mount point doesn't exist at all, it doesn't matter if the path is in a policy it will still give 403.

So for that reason I also changed the "inexistent" path to use the existing mount point, with a non-existent sub-path, rather than a mount point that didn't exist.

plugins/lookup/vault_list.py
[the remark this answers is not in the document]
briantist: it looks like https://github.com/ansible-collections/community.hashi_vault/commit/857568ab7e72c9826cf56b5bb3691a4d950433d3 only updated the module and not the lookup

plugins/lookup/vault_list.py
briantist: it looks like https://github.com/ansible-collections/community.hashi_vault/commit/857568ab7e72c9826cf56b5bb3691a4d950433d3 only updated the module and not the lookup

plugins/lookup/vault_list.py
briantist: it looks like https://github.com/ansible-collections/community.hashi_vault/commit/857568ab7e72c9826cf56b5bb3691a4d950433d3 only updated the module and not the lookup

plugins/lookup/vault_list.py
briantist: it looks like https://github.com/ansible-collections/community.hashi_vault/commit/857568ab7e72c9826cf56b5bb3691a4d950433d3 only updated the module and not the lookup

plugins/lookup/vault_list.py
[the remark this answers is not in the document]
tomkivlin: You're right, sorry - now updated in https://github.com/ansible-collections/community.hashi_vault/pull/343/commits/4050a8e2375e5595bdda227e95c8f1bc15983ac5

tests/integration/targets/lookup_vault_list/tasks/lookup_vault_list_test.yml
tomkivlin: Interesting.  When I test it against my vault server, for any non-existent path - whether there is a policy that covers it or not, I get a 403 on list operations.  For get etc. it behaves as you describe, but list appears to give 403 regardless of what the policy says.  e.g. I just created a user with no policy, and tried listing the secrets at the fakesecret mount point:

```
curl -v --header "X-Vault-Token: <redacted>" --request LIST http://127.0.0.1:8200/v1/fakesecret/metadata | jq .
*   Trying 127.0.0.1:8200...
* TCP_NODELAY set
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to 127.0.0.1 (127.0.0.1) port 8200 (#0)
> LIST /v1/fakesecret/metadata HTTP/1.1
> Host: 127.0.0.1:8200
> User-Agent: curl/7.68.0
> Accept: */*
> X-Vault-Token: <redacted>
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 403 Forbidden
< Cache-Control: no-store
< Content-Type: application/json
< Strict-Transport-Security: max-age=31536000; includeSubDomains
< Date: Tue, 17 Jan 2023 15:33:08 GMT
< Content-Length: 60
<
{ [60 bytes data]
100    60  100    60    0     0  20000      0 --:--:-- --:--:-- --:--:-- 20000
* Connection #0 to host 127.0.0.1 left intact
{
  "errors": [
    "1 error occurred:\n\t* permission denied\n\n"
  ]
}
```

When I run the integration tests with the expected error being "doesn't seem to exist" it fails, as the error that is returned by the API call / hvac module is "Permission denied".

```
TASK [lookup_vault_list : Failure expected when inexistent path is listed] *****
fatal: [testhost]: FAILED! => {"msg": "An unhandled exception occurred while templating '{{ lookup('community.hashi_vault.vault_list', vault_kv2_api_list_inexistent_path, **kwargs) }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while running the lookup plugin 'community.hashi_vault.vault_list'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Forbidden: Permission Denied to path 'kv2/metadata/__inexistant'.. Forbidden: Permission Denied to path 'kv2/metadata/__inexistant'."}
...ignoring

TASK [lookup_vault_list : assert] **********************************************
fatal: [testhost]: FAILED! => {
    "assertion": "test_inexistent.msg is search(\"doesn't seem to exist\")",
    "changed": false,
    "evaluated_to": false,
    "msg": "Expected failure but got success or wrong failure message."
}
```

I have removed this section from the policy, and get the same result.

```
  path "{{ vault_kv2_api_list_inexistent_path }}" {
    capabilities = ["list"]
  }
```

So I guess the question is - should we "fix" the Vault API response (i.e. raise an issue there), or modify the tests to expect what is provided (403), or both?  Or maybe I've got the wrong end of the stick!

tomkivlin: > So I guess the question is - should we "fix" the Vault API response (i.e. raise an issue there), or modify the tests to expect what is provided (403), or both? Or maybe I've got the wrong end of the stick!

Although, the tests run here provide the correct error, I don't really understand why.  I'm running Vault v1.9.2, I wonder if that could be an issue?  I'll upgrade and see...

```
TASK [lookup_vault_list : Failure expected when inexistent path is listed] *****
fatal: [testhost]: FAILED! => {"msg": "An unhandled exception occurred while templating '{{ lookup('community.hashi_vault.vault_list', vault_kv2_api_list_inexistent_path, **kwargs) }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while running the lookup plugin 'community.hashi_vault.vault_list'. Error was a <class 'ansible.errors.AnsibleError'>, original message: The path 'kv2/metadata/__inexistant' doesn't seem to exist.. The path 'kv2/metadata/__inexistant' doesn't seem to exist."}
...ignoring

TASK [lookup_vault_list : assert] **********************************************
ok: [testhost] => {
    "changed": false,
    "msg": "All assertions passed"
}
```

briantist: Interesting, it could be a version issue. I found this but it seems pretty old; I'd think this was updated before 1.9.x:
- https://github.com/hashicorp/vault/issues/1365

But yeah, since some versions respond with 404, we should account for those too. 

Even if some versions give a 403, that is already covered by the other 403 tests.

briantist: Just to confirm, in your local test to `http://127.0.0.1:8200/v1/fakesecret/metadata` was `fakesecret` actually created as a mount point? That part is key, it will continue to give 403 if the mount doesn't exist

tomkivlin: Ah ok, no - it doesn't.  That'll be why, thanks.

However, I just noticed that the test is testing a list operation for 'mount_point/metadata/path' which will never work anyway, even if the mount point and the path exist. (see https://github.com/ansible-collections/community.hashi_vault/pull/343#discussion_r1071924966).

tests/integration/targets/setup_vault_configure/vars/main.yml
```diff
@@ -18,6 +19,12 @@ vault_kv2_multi_api_path: '{{ vault_kv2_mount_point }}/data/{{ vault_kv2_multi_p
 vault_kv2_versioned_api_path: '{{ vault_kv2_mount_point }}/data/{{ vault_kv2_versioned_path }}'
 vault_kv2_delete_api_path: '{{ vault_kv2_mount_point }}/delete/{{ vault_kv2_versioned_path }}'
 vault_kv2_metadata_api_path: '{{ vault_kv2_mount_point }}/metadata/{{ vault_kv2_versioned_path }}'
+vault_kv2_api_list_path: '{{ vault_kv2_mount_point }}/metadata'
+
+vault_policy_api_list_path: 'sys/policies/acl'
+
+vault_kv2_api_list_inexistent_path: '{{ vault_kv2_mount_point }}/metadata/__inexistant'
```
briantist, line 26: typo
```suggestion
vault_kv2_api_list_inexistent_path: '{{ vault_kv2_mount_point }}/metadata/__inexistent'
```

tests/integration/targets/lookup_vault_list/tasks/lookup_vault_list_test.yml
tomkivlin: OK so whether it was updating or not, I get the (I think) expected behaviour now:

List works for:
* mount_point/metadata/path (where path is a path that contains secrets, not a secret name)
* mount_point/metadata

List gives 403 / permission denied for:
* mount_point_non_existent/metadata
* mount_point/metadata/path_non_existent (only if that is listed in a policy with "deny" capability)

List gives 404 / not found for:
* mount_point/metadata/path_non_existent

Just checking the tests again, and we were testing vault_list for:
* `'{{ vault_kv2_mount_point }}/metadata'` (expecting success)
* `'{{ vault_kv2_mount_point }}/metadata/__inexistent'` (expecting 404)

Which doesn't cover all the scenarios above.  I will make sure we have failure tests for:
* `'{{ vault_kv2_mount_point }}_inexistent/metadata/__inexistent'` (expecting 403)
* `'{{ vault_kv2_mount_point }}/metadata/__inexistent'` with explicit policy denying access (expecting 403)

And also will add a test for:
* `'{{ vault_kv2_mount_point }}/metadata/{{ vault_kv2_path }}'` (expecting success)

I'll try and make the examples clearer to pick up these nuances as well.  Thanks for your patience!

tomkivlin: Have done this in https://github.com/ansible-collections/community.hashi_vault/pull/343/commits/c294b877a9fe33ca872d5addfff1ef9ce89ef181

tests/integration/targets/lookup_vault_list/tasks/lookup_vault_list_test.yml
```diff
@@ -0,0 +1,137 @@
...
+      register: test_inexistent
+      ignore_errors: true
+
+    - assert:
+        that:
+          - test_inexistent is failed
+          - test_inexistent.msg is search("doesn't seem to exist")
+        fail_msg: "Expected failure but got success or wrong failure message."
+
+    # If an inexistent path is listed by is included in a policy statement that denies access, the API returns a 403 error.
```
briantist, line 105: nit/typo
```suggestion
    # If an inexistent path is included in a policy statement that denies access, the list API returns a 403 error.
```

tests/integration/targets/module_vault_list/tasks/module_vault_list_test.yml
```diff
@@ -0,0 +1,100 @@
...
+        path: "{{ vault_kv2_api_list_inexistent_path }}"
+      ignore_errors: true
+
+    - assert:
+        that:
+          - test_inexistent is failed
+          - test_inexistent.msg is search("doesn't seem to exist")
+        fail_msg: "Expected failure but got success or wrong failure message."
+
+    # If an inexistent path is listed by is included in a policy statement that denies access, the API returns a 403 error.
```
briantist, line 89: nit/typo
```suggestion
    # If an inexistent path is included in a policy statement that denies access, the list API returns a 403 error.
```

tests/integration/targets/module_vault_list/tasks/module_vault_list_test.yml
```diff
@@ -0,0 +1,100 @@
...
+      register: test_inexistent_unauthorized
+      community.hashi_vault.vault_list:
+        path: "{{ vault_kv2_api_list_inexistent_unauthorized_path }}"
+      ignore_errors: true
+
+    - assert:
+        that:
+          - test_inexistent_unauthorized is failed
+          - test_inexistent_unauthorized.msg is search("Permission Denied")
+        fail_msg: "Expected failure but got success or wrong failure message."
```
briantist, line 100: nit: eol
```suggestion
        fail_msg: "Expected failure but got success or wrong failure message."

```

tests/unit/fixtures/kv2_list_response.json
```diff
@@ -0,0 +1,15 @@
...
+            "Secret2"
+        ]
+    },
+    "lease_duration": 0,
+    "lease_id": "",
+    "renewable": false,
+    "request_id": "02e4b52a-23b1-9a1c-cf2b-3799edb17fed",
+    "warnings": null,
+    "wrap_info": null
+}
```
briantist, line 15: nit: eol
```suggestion
}

```

tests/unit/fixtures/policy_list_response.json
```diff
@@ -0,0 +1,15 @@
...
+            "Policy2"
+        ]
+    },
+    "lease_duration": 0,
+    "lease_id": "",
+    "renewable": false,
+    "request_id": "96f2857e-5e33-1957-ea7e-be58f483faa3",
+    "warnings": null,
+    "wrap_info": null
+}
```
briantist, line 15: nit: eol
```suggestion
}

```

tests/unit/fixtures/userpass_list_response.json
```diff
@@ -0,0 +1,15 @@
...
+            "User2"
+        ]
+    },
+    "lease_duration": 0,
+    "lease_id": "",
+    "renewable": false,
+    "request_id": "8b18a5ca-9baf-eb7c-18a6-11be81ed95a6",
+    "warnings": null,
+    "wrap_info": null
+}
```
briantist, line 15: nit: eol
```suggestion
}

```

tests/integration/targets/setup_vault_configure/vars/main.yml
```diff
@@ -18,6 +19,15 @@ vault_kv2_multi_api_path: '{{ vault_kv2_mount_point }}/data/{{ vault_kv2_multi_p
...
 vault_kv2_delete_api_path: '{{ vault_kv2_mount_point }}/delete/{{ vault_kv2_versioned_path }}'
 vault_kv2_metadata_api_path: '{{ vault_kv2_mount_point }}/metadata/{{ vault_kv2_versioned_path }}'
+vault_kv2_api_list_mount_point: '{{ vault_kv2_mount_point }}/metadata'
+vault_kv2_api_list_path: '{{ vault_kv2_mount_point }}/metadata/{{ vault_kv2_path }}'
+
+vault_policy_api_list_path: 'sys/policies/acl'
+
+vault_kv2_api_list_inexistent_path: '{{ vault_kv2_mount_point }}/metadata/__inexistent'
+vault_kv2_api_list_inexistent_mount_point: '{{ vault_kv2_mount_point }}__inexistent/metadata'
+vault_kv2_api_list_inexistent_unauthorized_path: '{{ vault_kv2_mount_point }}/metadate/__inexistent_no_auth'
```
briantist, line 29: everything else looks good but I'm confused by this change in https://github.com/ansible-collections/community.hashi_vault/pull/343/commits/dd7c6779f1d41ad7b4d214d843ced0adbe04ee8a

briantist (commented):
thanks! a few very small things, otherwise it's really just the merge resolution commit that I'm confused about

I really appreciate the extra tests and attention to detail on the scenarios!

tests/integration/targets/setup_vault_configure/vars/main.yml
tomkivlin: Yeah me too, not sure what that's about..

tomkivlin, line 29:
```suggestion
vault_kv2_api_list_inexistent_unauthorized_path: '{{ vault_kv2_mount_point }}/metadata/__inexistent_no_auth'
```

tomkivlin: Have updated in https://github.com/ansible-collections/community.hashi_vault/pull/343/commits/2b43f4c1b7c7972a9ee15a47c4947a26df590418

briantist (approved):

briantist:
Thanks very much @tomkivlin ! Great work, I hope you'll stick around consider contributing more in the future :)

I'll look to get version `4.1.0` released soon.

tomkivlin:
> Thanks very much @tomkivlin ! Great work, I hope you'll stick and around consider contributing more in the future :)
> 
> I'll look to get version `4.1.0` released soon.

Me too, thanks for your support 👍

PR #348, opened 2023-02-22
Add `vault_kv2_write` module

### LINKED ISSUE #331: ✨ Feature Request: Dedicated Module for KV secrets
##### SUMMARY

I've read through some of the issues and understand the scope of the modules and that the vault_write Module has it's particular scope and purpose. It would be nice to have a module specialized on KV secrets, with a certain behavior that makes it more suitable for some ansible tasks than the vault_write module.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

Not defined yet

##### ADDITIONAL INFORMATION

This module should have the following functionality to set it apart from existing modules:
* Idempotence
* Different behaviours like
  * Update a secret (overwrite it, if content would change)
  * Append to a secret, if a new key is added for example
* Support features from KV Engine 2
  * Versioning
  * etc.

There is an existing module which implements this in quite a nice manner regarding the user experience. [Terry Howe Module](https://github.com/TerryHowe/ansible-modules-hashivault/blob/main/ansible/modules/hashivault/hashivault_secret.py), however, there are a few reasons why I wish to see such functionality in this collection!

devon-mar:
##### SUMMARY

Fixes #331 

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME
`vault_kv2_write`


briantist:
@devon-mar welcome back and thanks for submitting this! I gave it a quick skim, I'll have to go over it in detail but a few small things I noticed so far:
- examples reference kv1_write
- short description should not end in `.`
- we may want to avoid the hvac patch method because it will do its own read to compare data, and since plugin already did that, it would add a redundant read
  - we might even want to offer an option for whether the module should attempt to read the current state or not:
    - this means we can't given accurate changed status (always report changed)
    - but this prevents unnecessary reads when you don't care about the current state
    - but this will make the conditionals a little more complex, since we do still need to read once on a patch
    - when hvac gets the ability to make the server-side-patch calls, this could change again

anyway most of that is just ideas, I need to think about it a little when I'm more awake.

Thanks again!

devon-mar:
> * we may want to avoid the hvac patch method because it will do its own read to compare data, and since plugin already did that, it would add a redundant read

Yeah I took a look at hvac and it also uses the read to get the version for `cas`, hence why it's not supported with patch in this module. I'm fine with dropping the functionality altogether or adding a warning to the documentation.

>   * we might even want to offer an option for whether the module should attempt to read the current state or not:

Yeah, I don't even need idempotency in my own use case haha. 
I think I could just do `current_data={}` if the user doesn't want a read.


Let me know what you think.

plugins/modules/vault_kv2_write.py
```diff
@@ -0,0 +1,224 @@
...
+    path = module.params.get("path")
+    cas = module.params.get("cas")
+    data = module.params.get("data")
+    patch = module.params.get("patch")
+
+    module.connection_options.process_connection_options()
+    client_args = module.connection_options.get_hvac_connection_options()
+    client = module.helper.get_vault_client(**client_args)
+
+    if cas and patch is True:
```
briantist, line 154:
```suggestion
    if cas is not None and patch is True:
```
A CAS value of 0 is valid, to perform a write only when the secret doesn't exist, so we don't want to allow `cas=0` to be used with `patch=True` presumably.

Let's also add tests for this condition.

plugins/modules/vault_kv2_write.py
```diff
@@ -0,0 +1,224 @@
...
+    else:
+        changed = current_data != data
+
+    if changed is True and module.check_mode is False:
+        args = {
+            "path": path,
+            "secret": data,
+            "mount_point": mount_point,
+        }
+        if cas:
```
briantist, line 200:
```suggestion
        if cas is not None:
```

plugins/modules/vault_kv2_write.py
```diff
@@ -0,0 +1,224 @@
...
+            args["cas"] = cas
+
+        try:
+            if patch is True:
+                raw = client.secrets.kv.v2.patch(**args)
+            else:
+                raw = client.secrets.kv.v2.create_or_update_secret(**args)
+        except hvac.exceptions.Forbidden:
```
briantist, line 208: I think we might need to account for a few other exception types.
I believe we'd get `InvalidRequest` (400) if we tried to update a secret where cas is required but we don't supply it.

I'm also wondering what the exception/status code is if we supply a CAS value and it's wrong (like we gave it 2, but the latest is 1 or 3), I don't know off-hand, but we should add those scenarios to the tests.

tests/integration/targets/module_vault_kv2_write/tasks/test.yml
```diff
@@ -0,0 +1,311 @@
...
+      register: result
+
+    - assert:
+        that:
+          - result is failed
+          - result.msg is search("Permission denied reading")
+
+    - name: Test write cas (check=True)
+      check_mode: true
+      community.hashi_vault.vault_kv2_write: &write_case
```
briantist, line 226:
```suggestion
      community.hashi_vault.vault_kv2_write: &write_cas
```

tests/integration/targets/module_vault_kv2_write/tasks/test.yml
```diff
@@ -0,0 +1,311 @@
...
+      register: ci_read
+
+    - assert:
+        that:
+          - result is changed
+          - ci_read.result.data.data == write_data2
+          - ci_read.result.data.metadata.version == 3
+
+    - name: Test write cas (check=False)
+      community.hashi_vault.vault_kv2_write: *write_case
```
briantist, line 244:
```suggestion
      community.hashi_vault.vault_kv2_write: *write_cas
```

tests/integration/targets/module_vault_kv2_write/tasks/test.yml
```diff
@@ -0,0 +1,311 @@
...
+      register: ci_read
+
+    - assert:
+        that:
+          - result is changed
+          - ci_read.result.data.data == write_data3
+          - ci_read.result.data.metadata.version == 4
+
+    - name: Test write cas (idempotency)
+      community.hashi_vault.vault_kv2_write: *write_case
```
briantist, line 258:
```suggestion
      community.hashi_vault.vault_kv2_write: *write_cas
```

briantist (commented):
As I look a little more closely, I think we'll definitely want an option that does straight writes without reading first (would not be compatible with Patch until `hvac` has server-side patch support).

In addition to merely allowing for fewer roundtrips, it would support a "write-only" scenario where a token has write access to a path but not read access.

I'm also looking at the `patch` parameter and wondering about changing it from a boolean to a string with choices. Although we can't support server-side patch just yet, I do think we will eventually, and because the permissions needed for the read-then-write method are different, we'll want to make that configurable probably, similar to how it is in the CLI: https://developer.hashicorp.com/vault/docs/secrets/kv/kv-v2

> The `vault kv patch` command also supports a `-method` flag which can be used to specify HTTP `PATCH` or read-then-write. The supported values are `patch` and `rw` for HTTP `PATCH` and read-then-write, respectively.

The choices could look like `none`, `auto`, `patch`, `rw` maybe (we would exclude `patch` method now).

It might be premature optimization, but I'm trying to avoid a scenario where we later want to do this and have to choose between 1) a breaking change on the `patch` option's type (will take time to release due to deprecation), or 2) adding a second `patch_method` option (cluttered options), or 3) converting `patch` to a `raw` option that takes either a method or a boolean (more complicated to maintain and test).

I suppose with option 2), we could deprecate the `patch` option and then remove it.

I'm also open to splitting direct writes and patches to two different modules which could help in handling the various option interactions and reduce branching, though I'm having trouble at the moment deciding if splitting in two is even worse.

Anyway, thinking out loud, please let me know your thoughts!

plugins/modules/vault_kv2_write.py
devon-mar: I'll add a test for CAS required.


Regarding a wrong cas value:
```
hvac.exceptions.InvalidRequest: check-and-set parameter did not match the current version, on post http://vault:8200/v1/kv2/data/testproject/write1
```

I have a test for it https://github.com/ansible-collections/community.hashi_vault/blob/f2d1c2fbedf819c7ec1fb39927bca6654111cb85/tests/integration/targets/module_vault_kv2_write/tasks/test.yml#L271-L283


Should I also catch `InvalidRequest`?

briantist: I believe we catch that too, yes.

briantist:
btw if you want to accept multiple suggestions in one commit, you can do so by looking at the suggestions from the `Files changed` tab, and then there will be an option to "Add suggestion to batch" so you can group similar changes together that way and save a lot of time waiting for CI :)

devon-mar:
> In addition to merely allowing for fewer roundtrips, it would support a "write-only" scenario where a token has write access to a path but not read access.

I've added the `read` option which defaults to `false`.

> The choices could look like `none`, `auto`, `patch`, `rw` maybe (we would exclude `patch` method now).

What would the `auto` option do? Choose the method based on `hvac` version?

> It might be premature optimization, but I'm trying to avoid a scenario where we later want to do this and have to choose between 1) a breaking change on the `patch` option's type (will take time to release due to deprecation), or 2) adding a second `patch_method` option (cluttered options), or 3) converting `patch` to a `raw` option that takes either a method or a boolean (more complicated to maintain and test).
> I suppose with option 2), we could deprecate the `patch` option and then remove it.

I wouldn't even mind another option, leaving patch functionality out together and waiting for a feature request to add it back in. That way we can get an actual use case for it (unless you have one).

> I'm also open to splitting direct writes and patches to two different modules which could help in handling the various option interactions and reduce branching, though I'm having trouble at the moment deciding if splitting in two is even worse.

Yeah, I'm on the fence about this one. On one hand, I can't think of a scenario of the top of my head where I'd like to control whether I'd like to write/patch dynamically (i.e. jinja2 template). On the other, write/patch seem closely related and many other Ansible modules group different functionality together (see `systemd`, `yum`, etc).

plugins/modules/vault_kv2_write.py
```diff
@@ -0,0 +1,253 @@
...
+  - For detailed requirements, see R(the collection requirements page,ansible_collections.community.hashi_vault.docsite.user_guide.requirements).
+extends_documentation_fragment:
+  - community.hashi_vault.attributes
+  - community.hashi_vault.attributes.action_group
+  - community.hashi_vault.connection
+  - community.hashi_vault.auth
+  - community.hashi_vault.engine_mount
+attributes:
+  check_mode:
+    support: full
```
devon-mar, line 30: Should this be `partial`?

tests/integration/targets/module_vault_kv2_write/tasks/cleanup.yml
```diff
@@ -0,0 +1,14 @@
...
+  - name: Ensure test secrets are deleted
+    vault_ci_kv2_destroy_all:
+      path: "{{ item.path }}"
+      mount_point: "{{ item.mount | default(vault_kv2_mount_point) }}"
+    loop:
+      - path: "{{ vault_kv2_path }}/write1"
+      - path: "{{ vault_kv2_path }}/readonly"
+      - path: "{{ vault_kv2_path }}/writeonly"
+      - path: "write1"
+        mount: "{{ cas_required_vault_kv2_mount_point }}"
```
briantist, line 14: I would prefer this step be moved to `setup.yml`, before the creation. The destroy should be idempotent so it shouldn't fail on non-existence, and this addresses a (very) minor edge case where a previous run failed without running the `always` block that invokes the cleanup, for whatever reason.

Mostly though, I'd just like all the steps to be together.

An example of another module that does this is `vault_kv2_delete`'s `setup.yml`.

plugins/modules/vault_kv2_write.py
briantist: Yes I think so, I'll also suggest an update to the `details`.

```suggestion
    support: partial
```

plugins/modules/vault_kv2_write.py
```diff
@@ -0,0 +1,253 @@
...
+    details:
+      - Supported if I(read) is C(true).
```
briantist, line 32:
```suggestion
      - If I(read) is C(true), full check mode functionality is supported.
      - If I(read) is C(false), the status will always be C(changed) but a write will not be performed in check mode.
```

briantist:
> > In addition to merely allowing for fewer roundtrips, it would support a "write-only" scenario where a token has write access to a path but not read access.
> 
> I've added the `read` option which defaults to `false`.

The implementation looks good thank you, I'm thinking about whether the option should have a slightly more specific name, like `read_existing` or `read_before_write`, both for clarity within the module, and reducing ambiguity between options in other modules and plugins (in case any introduce a similar option). 

Other than that, I might want to make some changes to this depending on whether we keep patch mode or not, and depending on the other questions about patch implementation.

> > The choices could look like `none`, `auto`, `patch`, `rw` maybe (we would exclude `patch` method now).
> 
> What would the `auto` option do? Choose the method based on `hvac` version?

It would first try the `patch` method (server-side), and if it failed, would try `rw`. This would mimic the way the CLI does it.
There's a chance we'll implement this same "auto" functionality into `hvac` in some way too (see https://github.com/hvac/hvac/issues/925), and then in this collection we'll have to consider also whether we also check `hvac` version and fallback or not.

> > It might be premature optimization, but I'm trying to avoid a scenario where we later want to do this and have to choose between 1) a breaking change on the `patch` option's type (will take time to release due to deprecation), or 2) adding a second `patch_method` option (cluttered options), or 3) converting `patch` to a `raw` option that takes either a method or a boolean (more complicated to maintain and test).
> > I suppose with option 2), we could deprecate the `patch` option and then remove it.
> 
> I wouldn't even mind another option, leaving patch functionality out together and waiting for a feature request to add it back in. That way we can get an actual use case for it (unless you have one).

Sure, it's totally fine to leave out patch for now, and that puts aside a bunch of the open questions for the time being, and we can at least get the direct write functionality out the door.

I'm not too worried about waiting for a feature request, I think it will be used if it's available, but I want to be respectful of your time and effort and if it's not functionality you need right now, it might not be worth the additional complication I'm adding to do it all at once. This will still be a very useful and welcome addition!

So let me know what you think, if you'd like to get patch in at the same time, I'll put time into trying to resolve these questions to unblock you (and can add some commits myself as time allows to hep it along). 

If you'd rather take it out for now and get this landed first, please go ahead with the necessary updates, and then I'll continue reviewing from there.

> > I'm also open to splitting direct writes and patches to two different modules which could help in handling the various option interactions and reduce branching, though I'm having trouble at the moment deciding if splitting in two is even worse.
> 
> Yeah, I'm on the fence about this one. On one hand, I can't think of a scenario of the top of my head where I'd like to control whether I'd like to write/patch dynamically (i.e. jinja2 template). On the other, write/patch seem closely related and many other Ansible modules group different functionality together (see `systemd`, `yum`, etc).

Yeah, if we opt not to include patch in this iteration, we can open a discussion about separating it or adding it in (though it can be difficult to get input).

devon-mar:
> The implementation looks good thank you, I'm thinking about whether the option should have a slightly more specific name, like read_existing or read_before_write, both for clarity within the module, and reducing ambiguity between options in other modules and plugins (in case any introduce a similar option).

I think `read_before_write` is the better of the two since we may be creating a secret (so there isn't an "existing" one to read).

>Yeah, if we opt not to include patch in this iteration, we can open a discussion about separating it or adding it in (though it can be difficult to get input).

Since patch needs some more planning I'll take it out of this PR. We can always add it back in another PR.

devon-mar:
@briantist

Let me know what you want to rename `read` to + any other changes.

plugins/modules/vault_kv2_write.py
```diff
@@ -0,0 +1,221 @@
...
+    data:
+      foo: bar
+
+- name: Write with check and set
+  community.hashi_vault.vault_kv2_write:
+    url: https://vault:8201
+    path: caspath
+    cas: 0
+    data:
+      foo: bar
```
briantist, line 76: Let's add an example where `cas` is not `0`, and another where it is `0`, labeled (via `name`) such that it explains that `0` means it will only succeed if the secret does not exist.

plugins/modules/vault_kv2_write.py
```diff
@@ -0,0 +1,221 @@
...
+"""
+
+import traceback
+
+from ansible.module_utils._text import to_native
+from ansible.module_utils.basic import missing_required_lib
+from ansible_collections.community.hashi_vault.plugins.module_utils._hashi_vault_common import (
+    HashiVaultValueError,
+)
+from ansible_collections.community.hashi_vault.plugins.module_utils._hashi_vault_module import (
```
briantist, line 107: this is very minor, but I'm starting to prefer relative imports, and I'm changing existing absolute imports to relative when changes are made to existing files. Let's update these to that form (I don't know offhand the right relative level sorry for not including the suggestion).

briantist (commented):
> I think `read_before_write` is the better of the two since we may be creating a secret (so there isn't an "existing" one to read).

ok let's go with `read_before_write` then, or I'm open to other suggestions, let me know if you have other ideas too

Thanks for sticking with this it's looking really great!

devon-mar:
@briantist I've applied the requested changes.

briantist:
Thank you I'm AFK right now so I'll look later but before I forget, please add this module to the action group in runtime.yml

plugins/modules/vault_kv2_write.py
```diff
@@ -0,0 +1,229 @@
...
+        module.fail_json(
+            msg=missing_required_lib("hvac"),
+            exception=HVAC_IMPORT_ERROR,
+        )
+
+    mount_point = module.params.get("engine_mount_point")
+    path = module.params.get("path")
+    cas = module.params.get("cas")
+    data = module.params.get("data")
+    read = module.params.get("read_before_write")
```
briantist, line 153: Let's change this `read` variable (and the other references to it) to also be `read_before_write`

plugins/modules/vault_kv2_write.py
```diff
@@ -0,0 +1,229 @@
...
+    type: int
+    description:
+      - Perform a check-and-set operation.
+  read_before_write:
+    type: bool
+    default: false
+    description:
+      - Read the secret first and write only when I(data) differs from the read data.
+      - Requires C(read) permission on the secret if C(true).
+      - If C(false), this module will always write to I(path).
```
briantist, line 59:
```suggestion
      - If C(false), this module will always write to I(path) when not in check mode.
```

briantist (commented):
Hi sorry for the delay here. This looks great! I have two small suggested changes and then I think that's it; I've looked at it thoroughly and don't expect any more changes after that. Looking forward to getting this merged

devon-mar:
@briantist changes applied!

briantist (approved):

briantist:
@devon-mar thanks again for contributing this module! It has been [released in `v4.2.0`](https://github.com/ansible-collections/community.hashi_vault/releases/tag/4.2.0)!

PR #350, opened 2023-02-26
Raise error on option duplication in term string

### LINKED ISSUE #349: `hashi_vault` lookup plugin behavior on option duplication
##### SUMMARY
`hashi_vault` lookup plugin allows defining option duplicates.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

`hashi_vault` lookup plugin

##### ANSIBLE VERSION

```
ansible [core 2.14.2]
  config file = None
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/ansible-vault-lookup/lib/python3.10/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/user/ansible-vault-lookup/bin/ansible
  python version = 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0] (/home/user/ansible-vault-lookup/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```
Collection            Version
--------------------- -------
community.hashi_vault 4.1.0 
```

##### CONFIGURATION

```
CONFIG_FILE() = None
```

##### STEPS TO REPRODUCE

Requirements: 
- Vault instance with secrets: `secret1: value1`, `secret2: value2`.
- Environment variable `VAULT_TOKEN`.

```yaml
- hosts: localhost
  gather_facts: false
  tasks:
    - debug:
        msg: >-
          {{ lookup('hashi_vault',
          'secret=secret/data/test:secret1 secret=secret/data/test:secret2
          url=http://127.0.0.1:8200/') }}
```

##### EXPECTED RESULTS

Option duplication error. 

##### ACTUAL RESULTS

Plugin returns second secret value.

```
PLAY [localhost] ***********************************************************************************************************************************************************

TASK [debug] ***************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "value2"
}

PLAY RECAP *****************************************************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

vladislav-sharapov:
##### SUMMARY

Fixes #349 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`hashi_vault` lookup plugin

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/plugin_utils/_hashi_vault_lookup_base.py
+++ b/plugins/plugin_utils/_hashi_vault_lookup_base.py
@@ -42,6 +42,9 @@ class HashiVaultLookupBase(HashiVaultPlugin, LookupBase):
                 else:
                     raise AnsibleError("%s lookup plugin needs key=value pairs, but received %s" % (plugin_name, term))
 
+            if key in param_dict:
+                raise AnsibleOptionsError("Duplicate key '%s' in the term string '%s'" % (key, term))
+
             param_dict[key] = value
 
         return param_dict


plugins/plugin_utils/_hashi_vault_lookup_base.py
```diff
@@ -42,6 +42,9 @@ def parse_kev_term(self, term, plugin_name, first_unqualified=None):
                 else:
                     raise AnsibleError("%s lookup plugin needs key=value pairs, but received %s" % (plugin_name, term))
 
+            if key in param_dict:
+                raise AnsibleOptionsError("Duplicate key '%s' in the term string '%s'" % (key, term))
```
briantist, line 46: I can't make a direct comment/suggestion on a line that didn't change, but you'll probably want to change line 15 to this:

```python
from ansible.errors import AnsibleError, AnsibleOptionsError
```

changelogs/fragments/350-raise-error-on-option-duplication-in-term-string.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - parse_kev_term() - function parsed the term string with option duplication. Function now raises error on option duplication in the term string.
```
briantist, line 2:
```suggestion
  - hashi_vault lookup - duplicate term-string-based options do not raise an error, leading to unexpected behavior (https://github.com/ansible-collections/community.hashi_vault/issues/349).
```

briantist (commented):
One thing I'm thinking about is that while I do consider this a bug, the fix is a breaking change. I may want to defer merging/releasing until the next major version of the collection, which should be roughly around May (so not _too_ long).

I'm going to think about that a bit and get some community input if I can.

If we defer it until v5, one thing we can do is to change this PR to warn about the duplicate instead of raising an exception, with the explanation that in v5 it will become a terminating error, and then we can make the hard change later.

So we have some options here. Thanks for putting this up!

plugins/plugin_utils/_hashi_vault_lookup_base.py
vladislav-sharapov: Sure, I've missed this one.

changelogs/fragments/350-raise-error-on-option-duplication-in-term-string.yml
vladislav-sharapov: From [docs](https://docs.ansible.com/ansible/devel/community/development_process.html#creating-a-changelog-fragment):
"Use past tense to describe the problem and present tense to describe the fix."
Is it fine to use present tense to describe bug? Also changelog doesn't contain fix description.

briantist: Thank you, you're correct! I had that wrong in my mind, let's leave your wording as is, but please do add the link back to the issue.

vladislav-sharapov: I've added issue link. If you want we can change wording.

briantist:
Hi @vladislav-sharapov , I've decided to go with making this a warning for now, and then making the actual breaking change in `v5.0.0` so that there's a bit of lead time.

I've pushed some commits to:
- make the change to a warning (kept the exception raising code commented out for later)
- add unit tests
- add commented out integration tests for later
- update the changelog fragment

Please take a look and let me know if that looks ok to you.

If you'd like to put up a PR for the v5 work too, that'd be fine (it won't merge until around May when we're gearing up to release that version), if not that's fine too, I'll be able to take care of that when it gets closer.

changelogs/fragments/350-raise-error-on-option-duplication-in-term-string.yml
briantist: Updated in https://github.com/ansible-collections/community.hashi_vault/pull/350/commits/37e738c93a5352eca7162122516aa21d1d3029b0

plugins/plugin_utils/_hashi_vault_lookup_base.py
```diff
@@ -42,6 +42,13 @@ def parse_kev_term(self, term, plugin_name, first_unqualified=None):
                 else:
                     raise AnsibleError("%s lookup plugin needs key=value pairs, but received %s" % (plugin_name, term))
 
+            if key in param_dict:
+                removed_in = '5.0.0'
+                msg = "Duplicate key '%s' in the term string '%s'." % (key, term)
+                display.deprecated(msg + "\nIn version %s of the collection, this will raise an exception." % (removed_in, ), removed_in)
```
vladislav-sharapov, line 48: I can't figure out why use this complex form `(removed_in,)`. `% removed_id` works the same why, doesn't it? I'd use this variant if you don't want to use f-string:
```python
deprecated("%s\nIn version %s of the collection, this will raise an exception." % (msg, removed_in), removed_in)
```
f-string variant:
```python
display.deprecated(f"{msg}\nIn version {removed_in} of the collection, this will raise an exception.", removed_in)
```

vladislav-sharapov:
> If you'd like to put up a PR for the v5 work too, that'd be fine (it won't merge until around May when we're gearing up to release that version), if not that's fine too, I'll be able to take care of that when it gets closer.

I'm not in a hurry. It's up to you when the PR should be merged.

plugins/plugin_utils/_hashi_vault_lookup_base.py
briantist: I got into the habit of using a tuple with a single value `(removed_in,)` because if `removed_in` happened to itself be a tuple, then the `% removed_in` form would fail if its contents don't match the number of templated items. 

It's mostly guarding against variables that could have different types; it's not important here I guess, since we're setting the variable explicitly, but it's a thing I try to always do anyway for consistency.

The reason we haven't used f-strings historically is that we had to support Python 2.7/3.5. That's no longer the case, but I keep using `%` strings for consistency.

I might look to do a bulk update at a later time, but not sure I want to start introducing different methods one at a time.

---

All that aside, this call will be gone in a few months for v5 😝

briantist:
> > If you'd like to put up a PR for the v5 work too, that'd be fine (it won't merge until around May when we're gearing up to release that version), if not that's fine too, I'll be able to take care of that when it gets closer.
> 
> I'm not in a hurry. It's up to you when the PR should be merged.

With this PR changing to a warning, it will merge as soon as it's ready, I think probably today as long as you're ok with the changes I put up.

We'll need a second PR for the v5 changes; that's the one that will have to wait until closer to that release. I'm fine to handle that one, or you can submit those changes as well (now or leading up to the v5 release) if you'd like to, either way all good.

vladislav-sharapov:
Thank you for clarifying. I'm OK with the changes. I think you have better understanding how the v5 changes should look like.

briantist (approved):

PR #391, opened 2023-07-22
update hvac error handling (Open todo)

mathijswesterhof:
##### SUMMARY

This pull request addresses the hashi_vault_common module. In specific, the handling of the HVAC import. In the current situation the import is tried and if failed the exception is handled by setting `HAS_HVAC` to false. this state is later not used to signal the missing lib to the user.
This pull request fixes this by raising an HVAC specific error which in turn is picked up by the modules that use the common module.
The reason for raising the custom exception is that the module and plugin utils both make use of this class and both have their own error handling service (being; plugins via AnsibleError and modules via the fail handler of the AnsibleModule class)

**Note that**
 - It is not perse a bugfix but more of an open todo still in the code.
 - fixing this todo opens up a lot of cleaning in the separate modules where this is being double checked.

I'm not sure if I can confidently remove all the HVAC import try-catch blocks, but based on this change it is a lot easier to write an adapter that can be used for all current modules and future modules significantly reducing duplicated code.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

 - module_utils/_hashi_vault_common.py
 - module_utils/_hashi_vault_module.py
 - plugin_utils/_hashi_vault_plugin.py

##### ADDITIONAL INFORMATION

 Implementing the following change will have a positive effect in replcated code, some examples below for some widely used modules: vault_login
afbeelding

vault_list
afbeelding

before
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'hvac'
fatal: [192.x.x.x]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (hvac) on ansible-box's Python /usr/bin/python3. Please read module documentation and install in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"}
```

after
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: No module named 'hvac'
fatal: [192.X.X.X]: FAILED! => { "changed": false, "msg": "Failed to import the required Python library (hvac) on ansible-box's Python /usr/bin/python3. Please read module documentation and install in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"}
```


mathijswesterhof:
Small note on this pr, I have no clue how to fully and properly run the pytest tests. I'm constantly getting the error that there's a relative import that's missing even though it's clearly there.
```
     from .compat import mock
E   ImportError: attempted relative import with no known parent package
```
I'm probably doing something simply stupid but if anyone can get me up to speed with that, that'll be amazing cause I want to properly test everything, especially with the new modules being build.

just for debugging purposes, I'm currently running `pytest -c tests/config.yml ./tests` in the root folder of the project

briantist:
> Small note on this pr, I have no clue how to fully and properly run the pytest tests. I'm constantly getting the error that there's a relative import that's missing even though it's clearly there.
> 
> ```
>      from .compat import mock
> E   ImportError: attempted relative import with no known parent package
> ```
> 
> I'm probably doing something simply stupid but if anyone can get me up to speed with that, that'll be amazing cause I want to properly test everything, especially with the new modules being build.
> 
> just for debugging purposes, I'm currently running `pytest -c tests/config.yml ./tests` in the root folder of the project

Hi @mathijswesterhof , welcome and thank you for putting up this PR.

In ansible collection, we don't really run `pytest` directly, but rather use `ansible-test` to invoke them. I highly recommend using that tool's container functionality (which is usually used with pre-built containers). The unit and sanity tests are the easiest ones to run locally on your machine.

The first step is to ensure that your checkout directory structure is correct. Your repository should be checked out in this structure: `<parent>/ansible_collections/community/hashi_vault`

This structure is important!

Then, from the root of the collection:

```
ansible-test sanity --docker default
```

```
ansible-test units --docker default
```

Units will run their tests across all supported python versions (within the container). For faster local testing you can choose a single python version like so:

```
ansible-test units --docker default --python 3.10
```

For more detailed information, please see our [Contributor guide](https://docs.ansible.com/ansible/devel/collections/community/hashi_vault/docsite/contributor_guide.html).

Let me know if you still need assistance running tests locally, or if the documentation there can be improved.

---

I will have to look over this PR more deeply when I have some more time. I'm traveling now so I will be slower to respond and review for the next several weeks, but I will at least try to approve the CI for new commits while I'm away. I strongly recommend getting set up to run the tests locally so that you can get much faster feedback on your changes.

Thank you!

mathijswesterhof:
@briantist I finally had some spare time to put into this again. sorry for the insane long delay. I added some base tests for  `_hashi_vault_common`  I tried to trigger code-coverage locally and it shows all tests succeed (but no new CodeCov percentages)

a few questions:
 - should i move the tests into a separate file? (I test another class, but it's in the same file so I assumed it should also go in the same test file)
 - how do I trigger the codeCov bot? (if I'm able to do that at all)
 - should I write an integration test for both the `_hashi_vault_plugin` and `_hashi_vault_module` files? (I assume integration since the init is already tested now and the only addition is the try block with the exception handover)

briantist:
> @briantist I finally had some spare time to put into this again. sorry for the insane long delay. I added some base tests for `_hashi_vault_common` I tried to trigger code-coverage locally and it shows all tests succeed (but no new CodeCov percentages)
> 
> * how do I trigger the codeCov bot? (if I'm able to do that at all)

I don't think you can trigger codecov locally because you'd need a token to upload coverage reports to it, but don't worry about that. To be honest, I never run coverage locally, I make sure the tests pass locally and then let CI do it with coverage so we can get the nice reports from codecov. I realize that's not so good when it's your first contribution and you have to wait for me to approve each run, you can do a lot from a single online coverage report since you can browse it and see all the missed lines so easily.

> * should i move the tests into a separate file? (I test another class, but it's in the same file so I assumed it should also go in the same test file)
> * should I write an integration test for both the `_hashi_vault_plugin` and `_hashi_vault_module` files? (I assume integration since the init is already tested now and the only addition is the try block with the exception handover)

I will have to defer these questions until I can give this a proper thorough review. I am incredibly short on time lately so I will try to get back to this as soon as I can.

mathijswesterhof:
@briantist any updates?

briantist:
> * should i move the tests into a separate file? (I test another class, but it's in the same file so I assumed it should also go in the same test file)

for now it's probably fine to keep them where they are, we can come back to that if it seems like they'd be better moved

> * should I write an integration test for both the `_hashi_vault_plugin` and `_hashi_vault_module` files? (I assume integration since the init is already tested now and the only addition is the try block with the exception handover)

I think they mostly get tested in integration implicitly via the integration tests for the plugins and modules that use them. The units help a lot with test cases that are hard to replicate in integration. `hvac` missing is one of those difficult test cases when it comes to ansible testing because `ansible-test` installs it for us and there isn't an easy way to tell it to do so selectively.

briantist:
It might be good to start migrating plugins and modules to use this method. Could you do one plugin and one module and push those up so that I can review? It would be good to (roughly) have each plugin and module be in its own commit, but not strictly necessary. Thanks again for your work on this, sorry for the delays.

mathijswesterhof:
Not sure why this one failed on the CI - LI, but main is also failing (https://github.com/ansible-collections/community.hashi_vault/actions/runs/9528624007)

Updated the KV modules and Lookups to use the `HashiVaultHelper` hvac entry instead of an import. (which mainly just removes the import code)

I have also removed the test that runs over the import try/except with the `HAS_HVAC` switch case. That test is now being done in the `HashiVaultHelper` test class.

I'm still doubting if it is neater to add:
```
    def get_hvac(self):
        return self.hvac
```
to the `HashiVaultHelper` Class and then use `hvac = module.helper.get_hvac()` to make it a bit more clear where HVAC is comming from. let me know if that is a good plan.

briantist:
> Not sure why this one failed on the CI - LI, but main is also failing (https://github.com/ansible-collections/community.hashi_vault/actions/runs/9528624007)

yeah the local integration tests have been failing, nothing to do with your changes, I need to get around to fixing that separately.

> I'm still doubting if it is neater to add:
> 
> ```
>     def get_hvac(self):
>         return self.hvac
> ```
> 
> to the `HashiVaultHelper` Class and then use `hvac = module.helper.get_hvac()` to make it a bit more clear where HVAC is comming from. let me know if that is a good plan.

I don't have a strong opinion (yet) but one way that method could be advantageous is in unit tests; it makes it easier to mock or wrap in the modules. As you said:

> I have also removed the test that runs over the import try/except with the `HAS_HVAC` switch case. That test is now being done in the `HashiVaultHelper` test class.

that makes sense, but with a method we could more easily add a test to enforce that modules are actually calling the method, like:

```python
get_hvac = mock.Mock(wraps=helper.get_hvac)
# ...
get_hvac.assert_called_once()
```

(something like that)

---

disclaimer: I haven't given it deep thought yet, feel free to play around with the different implementations

mathijswesterhof:
I've adjusted the function a bit, since the only thing that is being called is within the exceptions subset. I personally like it better since it's more clear what it is for now. I'll also look into writing the suggested tests with the assertions probably tomorrow.

briantist (commented):
I've just thought of a possible downside of wrapping this in a function: it will probably ruin intellisense in code editors 😔 

```python
hvac = self.get_hvac()
```

If we try to do `hvac.<wait for pop-up here>` I think it will not work, which is unfortunate.
I can't think of any type-hinting that would fix that either since we need to tell it that it returns a _specific_ module (`hvac`) and not a module type... I haven't loaded any of this in an editor yet, can you confirm? Also does using `self.hvac` by itself work with intellisense (outside of the helper class)?

---

btw I fixed the local integration tests, if you rebase or merge `main` into your branch, you should see the CI errors go away.
#441

plugins/module_utils/_hashi_vault_common.py
```diff
@@ -17,22 +17,29 @@
...
-        pass
+        try:
+            import hvac
+            self.hvac = hvac
+        except ImportError as e:
+            from ansible.module_utils.basic import missing_required_lib
+            raise HashiVaultHVACError(error=str(e), msg=missing_required_lib('hvac'))
+
+    def get_hvac_exceptions(self):
+        return self.hvac.exceptions
```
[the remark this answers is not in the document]
mathijswesterhof: On one side that was also my first thought. But diving a bit deeper into the code I noticed that only three modules (excluding tests) require the Hvac import.

`vault_pki_generate_certificate.py`, `_auth_method_token.py` and  `_hashi_vault_common.py`. The last is the main player, having the setup for the client and thus also the potential `get_hvac` function. The first only needs Hvac for a constant and the second is a bit tricky with exceptions as well.

My point being that almost all Hvac related things are as good as already handled in the  `_hashi_vault_common.py`. and keeping them there would make it nice and clean. the only thing that's supposedly out of bounds there are the exceptions because they need to be handled on the module level with the json reports.

The trade-off;
having `get_hvac().exceptions` and then the underlying exceptions would have an extra level of mocking potentially. Over, having more possibility to in the future use other sub-categories of hvac.

In the end I have a slight preference over the exceptions function but implementation wise I can do both.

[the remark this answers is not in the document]
mathijswesterhof: @briantist I've quickly looked into the intellisense situation. completion works for the import statement. but from that point (either directly as `hvac.` or `self.hvac.`) it only shows `v1` or `client` and I think that that has to do with the `__init__.py` file. The `hvac.exceptions` are not detected either  way.

afbeelding
Including a python type-hint also does not seem to work even though it seems to detect it as a package. but this brings the extra problem that hvac needs to be imported from outside the function and therefor also needs to have it's 'is installed' tests outside the class. so that resolves to a no-go.

On that note, I did find another way to go over all the exceptions without repeating them in every module.
It might not be the cleanest way but it does tick the box of keeping the hvac module in the `_hashi_vault_common.py`

I've put in two exceptions for now as a trial
afbeelding

This would be the three-way comparison where the first is the original, the second is the `get_hvac().exceptions` or `get_hvac_exceptions()` (both would have virtually identical implementation) and the last would be with a wrapper. 
afbeelding

Curious about your opinion and final decision implementation wise. Also would be interested in a short overview of what still needs to be done before it is merge-able.

mathijswesterhof:
@briantist I've reverted the exceptions function and implemented the `get_hvac().exceptions` like you suggested. I've also did this for all implementations i could find. Currently the pipeline is failing on Devel and after some digging it looks like it failed in a function inside the Ansible base code So I'm not sure if it will get fixed after a few commits or if that is something we should work around. I'll keep tabs on that this week. I want to get started with the new ansible functions next week so if I can get a review before then that would be amazing. otherwise it'll have to wait untill mid september

briantist:
@mathijswesterhof the devel failures are not related to this, that needs to be fixed separately for the repo. 

> I want to get started with the new ansible functions next week so if I can get a review before then that would be amazing. otherwise it'll have to wait untill mid september

Unfortunately I don't think it'll be possible. It's an incredibly busy time for me and I have far less time for this than I have in the past, hoping it eases up but looking at my week ahead it feels unlikely I'll be able to take a look before then, we'll see though.

mathijswesterhof:
@briantist any update?

briantist:
The CI for the repo has been fixed, so let's get a new CI run after you update your branch

mathijswesterhof:
@briantist it seems like all checks are successful (apart from the codecov upload timeout) and there are no instances of the import anymore.

mathijswesterhof:
> Should I create a separate issue for this?

https://github.com/ansible-collections/community.hashi_vault/pull/455
It was a fairly easy fix, so I've created a pull request just in case; to speed up the process

mathijswesterhof:
@briantist checks have re-run and everything is green again :D

mathijswesterhof:
@briantist can you give me an update on roughly when you can review this? then i can get going with the implementation of the policies and roles routes.

plugins/module_utils/_hashi_vault_module.py
```diff
@@ -14,10 +14,7 @@
 
 from ansible.module_utils.basic import AnsibleModule
 
-from ansible_collections.community.hashi_vault.plugins.module_utils._hashi_vault_common import (
-    HashiVaultHelper,
-    HashiVaultOptionAdapter,
-)
+from ansible_collections.community.hashi_vault.plugins.module_utils._hashi_vault_common import HashiVaultHelper, HashiVaultHVACError, HashiVaultOptionAdapter
```
briantist, line 17: nit: please revert back to the multi-line style for this, it makes it easier to see when the list was changed in git history

briantist, line 17:
```suggestion
from ansible_collections.community.hashi_vault.plugins.module_utils._hashi_vault_common import  (
    HashiVaultHelper,
    HashiVaultHVACError,
    HashiVaultOptionAdapter,
)
```

plugins/module_utils/_hashi_vault_module.py
```diff
@@ -14,8 +14,9 @@
 
 from ansible.module_utils.basic import AnsibleModule
 
-from ansible_collections.community.hashi_vault.plugins.module_utils._hashi_vault_common import (
+from ansible_collections.community.hashi_vault.plugins.module_utils._hashi_vault_common import  (
```
briantist, line 17:
```suggestion
from ansible_collections.community.hashi_vault.plugins.module_utils._hashi_vault_common import (
```

briantist (approved):
@mathijswesterhof thanks very much for your work on this, sticking with it, and your patience with my dwindling availability. It looks great and I really appreciate your work! 

Re: my previous requested change I went ahead and just committed that myself, please review and be sure to pull on your end if you'll be making any other changes. If it all looks good to you then we can merge (as long as my change didn't mess up the tests).

mathijswesterhof:
@briantist I agree with the changes you've made, I think I messed this up along the way with the various experiments I did on making it look nice and trying to optimize the different components.

loelkes:
Is there already an estimated date for a release including this? @briantist

briantist:
@loelkes there is not an estimated date right now, however this change is not user facing so it should not change your usage of the collection. Is there something specific you're looking for in the release?

loelkes:
> so it should not change your usage of the collection

I got an import error using the collection in a Python3.11 environment, including the `hvac` package. The problem has somehow been fixed in this release, as the import "tests" have been removed. The other solution was to install `hvac` globally, outside the virtual environment, which doesn't work for me.

Using the collection from `main` (which contains this fix) works as expected. Maybe I'm missing something? I found it quite odd and will check again and post some details here about where and in which environments the bug occurred.

briantist:
thank you @loelkes please do post any info you have. The collection is tested against python 3.11 so there shouldn't be any issue with that, but I am curious what's going on there. It would probably be best as a new [Q&A Discussion Topic](https://github.com/ansible-collections/community.hashi_vault/discussions/categories/q-a) unless it can definitively be tied to this PR though.

Issue #446, opened 2024-08-02
vault_kv2_delete: Add posibility to destroy


Kloppi313:
##### SUMMARY
It is not possible to destroy something in vault.
I think a boolean switch would be great:

Adding something like the parameter `destroy: True/False` with standard false would do the trick.

The context is the wish to [destroy an entire secret with all versions](https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v2#delete-metadata-and-all-versions) while decomissining a host with its saved secrets.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
vault_kv2_delete

##### ADDITIONAL INFORMATION

```yaml
- name: Delete all of the secret/mysecret secret.
  community.hashi_vault.vault_kv2_delete:
    url: https://vault:8201
    path: secret/metadata/mysecret
    auth_method: userpass
    username: user
    password: '{{ passwd }}'
    destroy: True
  register: result
```

briantist:
Hi @Kloppi313 , destroy was discussed a little bit in the PR for `vault_kv2_delete`
- https://github.com/ansible-collections/community.hashi_vault/pull/304

I think I'd rather see this as its own module, `vault_kv2_destroy`. The reason is mainly that it is a different functionality from deletes, so it should be separate, and it will be easier to test each module on its own that way.

It's currently waiting for a contributor to implement it though.

Kloppi313:
Hi @briantist,

so would you prefer renaming this issue?
Wouldn't it be a good start to duplicate vault_kv2_delete module and changing everything to destroy for a naive start?

briantist:
Sure, I've renamed this issue. Starting from `vault_kv2_delete` is fine, we don't have a "template" so I usually start with some existing content when creating a new one anyway. If you'd like to take this on please feel free to submit a PR!

Kloppi313:
I created a PR, but it is not what I firstly had in mind -> removing a complete endpoint, not only to destroy versions. But hey, it is a new feature.

octavian2204:
maybe instead of creating a separate module would be wiser to add `state: present/absent/destroy` to existing ones?

cyrillesondag:
I'm also interested in this feature — let me know if I can help.
I agree with you, @octavian2204 one module with different states will be much easier to use.

PR #449, opened 2024-08-16
added a vault_kv2_delete with destroy optione

Kloppi313:
##### SUMMARY

The `vault_kv2_delete` module can only delete data. There is little code change to also destroy. It seems more reasonable to add the option instead of creating a new module

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

vault_kv2_delete

##### ADDITIONAL INFORMATION

Idea came up here #446.

```paste below

```

### DIFF
--- a/plugins/modules/vault_kv2_delete.py
+++ b/plugins/modules/vault_kv2_delete.py
@@ -40,6 +40,13 @@ extends_documentation_fragment:
   - community.hashi_vault.auth
   - community.hashi_vault.engine_mount
 options:
+  destroy:
+    description:
+      - Destroy credential instead of delete
+      - Versions is required when used
+    type: bool
+    default: False
+    required: False
   engine_mount_point:
     default: secret
   path:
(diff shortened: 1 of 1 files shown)


briantist:
note: there are collection-wide CI failures at the moment that I need to deal with that aren't related to your PR:
- https://github.com/ansible-collections/community.hashi_vault/actions/runs/10421215850

---

While I will consider this change rather than a dedicated destroy module, I do think some of the reasons for having a separate module still stand: namely around testing and option combinations (like versions being required now but only in certain circumstances). Maybe they aren't as severe as I thought either though.

I'll point out that coverage for the existing module is 100%:
- https://app.codecov.io/gh/ansible-collections/community.hashi_vault/blob/main/plugins%2Fmodules%2Fvault_kv2_delete.py

I will not want that bring that lower, and this PR will need both integration and unit test additions. The process of adding those should begin highlight how much complexity (or not) this change ends up adding, vs. having a dedicated module, so let's see how adding the tests play out.

(and of course: coverage number is not the end-all-be-all: 100% coverage doesn't mean we're testing all the right things, but <100% does mean we are missing some code paths; I will try to point out any test cases I think are important when I have proper time to review again)

Kloppi313:
Just my 2cents why I did it within this module:

- I am no programmer, I am sysadmin, so the copycat approach is easier for me
- no code duplication with a new module
- I think I would be maintainer of this new module, because I am no programmer, that might be ne good choice
- a new module would also need tests(?)

Please be advised: the case `destroy: true && versions not defined` is not covered within the if/else branch, as I not know how to exit correctly for ansible. This should be a finding of testing

PR #467, opened 2025-07-13
Tests/fix various issues 2.19

briantist:
##### SUMMARY
Fix tests to work with Ansible 2.19

Closes: #464 

##### ISSUE TYPE

- Tests/CI Pull Request

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -40,11 +40,7 @@ jobs:
         test_container:
           - default
         ansible:
-          - stable-2.14
-          - stable-2.15
-          - stable-2.16
-          - stable-2.17
-          - stable-2.18
+          - stable-2.19
           - devel
     steps:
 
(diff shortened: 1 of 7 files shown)


mariolenz:
I think it would be a good idea to add ansible-core 2.19 to .github/workflows/ansible-test.yml since devel got bumped to 2.20 recently. Since it works with devel, there shouldn't be any problems there.

Sooner or later, you should do it anyway. So why not do it now and save the time to do an extra PR for this :smirk:

briantist:
Thanks @mariolenz , I will see if I can squeeze out a new major release in that timeframe and try to cut all those old versions.

In the meantime, I've added 2.19 to sanity tests in this PR and removed all the other versions except devel. I don't think there's much value in sanity testing for anything other than the latest stable and devel.

mariolenz:
> I don't think there's much value in sanity testing for anything other than the latest stable and devel.

In theory, I think you should test with all ansible-core versions you support officially. But I agree with you there... in practice, if the CI is happy with the latest stable and devel, the tests also succeed with older releases in my experience :-)

